PR #15, opened 2020-05-08
add cert-manager to andes, antu, kueyen and ruka

hreinking:


andes/cert-manager/cm-and-clusterissuer.sh
```diff
@@ -0,0 +1,45 @@
...
+kubectl apply -f https://raw.githubusercontent.com/jetstack/cert-manager/release-0.11/deploy/manifests/00-crds.yaml --validate=false
+helm install cert-manager -n cert-manager jetstack/cert-manager
+cat > secret.yaml << END
+apiVersion: v1
+kind: Secret
+metadata:
+  name: aws-route53
+  namespace: cert-manager
+data:
+  aws_key: $(AWS_SECRET_KEY)
```
adrienthebo, line 16: Is this meant to perform variable expansion? If so then it should be `${AWS_SECRET_KEY}`.

In addition we should add `set -u` so the script aborts if this variable is unset.

hreinking: The "set -u" is a wonderful idea. And yes, the AWS_SECRET_KEY must be set as an environmental variable before running the script.

PR #18, opened 2020-05-27
updates to ruka

hreinking:
adding ruka04 and comment out ruka01 until further notice


ruka/metallb/mlb-l2.yaml
```diff
@@ -10,4 +10,4 @@ data:
     - name: default
       protocol: layer2
       addresses:
-      - 139.229.135.150-139.229.135.249
+      - 139.229.135.150-139.229.135.199
```
jhoblitt, line 13: It looks like the original IP range is still in IPAM (the DHCP pool seems to be missing). The revised range also appears to overlap with the DHCP pool: https://github.com/lsst-it/lsst-itconf/blob/master/hieradata/site/ls/role/foreman.yaml#L29 and one of them needs to be adjusted.  The DHCP pool is probably much larger than is necessary so it may make more sense to shrink it.

ruka/rke/cluster.yml
```diff
@@ -33,22 +49,22 @@ nodes:
...
+  hostname_override: "ruka04"
   user: rke
   docker_socket: /var/run/docker.sock
   ssh_key: ""
   ssh_key_path: ~/.ssh/id_rsa
   ssh_cert: ""
   ssh_cert_path: ""
   labels:
-    role: storage-node
+    role: storage-node
```
jhoblitt, line 67: Unnessicary trailing whitespace.

ruka/rook-ceph/rook-ceph-node-cleanup.sh
```diff
@@ -6,7 +6,7 @@ SSH_USER=jhoblitt_b
 
 for n in $(seq 1 3);
```
jhoblitt, line 7: This is special use script but it may make sense to adjust the range to 2-4 here just in case someone forgets and tries to run it as-is.

ruka/rke/cluster.yml
hreinking: Whitespace deleted

ruka/rook-ceph/rook-ceph-node-cleanup.sh
hreinking: range updated

ruka/metallb/mlb-l2.yaml
hreinking: This one will need more digging. I'll create another branch for both repos once I figure out a simple solution

PR #21, opened 2020-05-28
It 2232 cert manager

hreinking:
Same as before but renamed the branch so it has a dash instead of a underscore

### DIFF
--- a/andes/cert-manager/cm-and-clusterissuer.sh
+++ b/andes/cert-manager/cm-and-clusterissuer.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -ex
+set -exu
 
 kubectl create ns cert-manager
 helm repo add jetstack https://charts.jetstack.io
--- a/andes/cert-manager/cm-and-clusterissuer.sh
+++ b/andes/cert-manager/cm-and-clusterissuer.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+set -ex
+
+kubectl create ns cert-manager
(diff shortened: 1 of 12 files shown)


andes/cert-manager/cm-and-clusterissuer.sh
```diff
@@ -0,0 +1,45 @@
...
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt
+  namespace: cert-manager
+spec:
+  acme:
+    server: https://acme-v02.api.letsencrypt.org/directory 
+    privateKeySecretRef:
+      name: letsencrypt
+    email: $(EMAIL)
```
adrienthebo, line 31: Bash variable expansion is `${VAR}`; shell command expansion is `$(command)`.

This is a hard blocker as this script cannot work until variable expansion is corrected.

ruka/cert-manager/cm-and-clusterissuer.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+set -ex
```
adrienthebo, line 3: Missing `set -u`.
```suggestion
set -exu
```

antu/cert-manager/cm-and-clusterissuer.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+set -ex
```
adrienthebo, line 3: Missing `set -u`

```suggestion
set -exu
```

PR #25, opened 2020-07-06
It 2268 vk8s

hreinking:


hreinking:
Due to IT-2380, just noticed that vk8s config was never merged. Doing it now so I can proceed with the ticket.

PR #91, opened 2022-08-03
(pillan) extend metallb pool to 140.252.146.46-140.252.146.59

jhoblitt:
This add +4 IPs to the pool. The 140.252.146.32/27 subnet is now
exhausted except for 3 IPs in the DHCP pool.


cbarria (approved):
extended range of subnet for kubernetes cluster from tucson 👍

PR #94, opened 2022-08-18
(tu site) chart updates related to k8s 1.23 migration

jhoblitt:
Strange issues were observed when restarting k8s in that the ingress
controller would not pass its own http health checks.  Attempting to
reinstall the same version failed to resolve the problem.  Updating to
4.2.1 fixed the issue. The root cause is unknown.


cbarria (approved):
saw it during the TTS upgrade

PR #96, opened 2022-08-23
add quotas to ceph pools on ruka/pillan/yagan

jhoblitt:
Quotas were already manually set on pillan & yagan and the values have been copied to the ceph CRDs without modification.


cbarria (approved):
much needed 👍

PR #100, opened 2022-08-29
synchronize rook/ceph/ceph-csi configuration across all clusters

jhoblitt:
The pillan configuration, which already included s3 & nfs, was copied to ruka for development and testing/updates. Rook was updated to 1.9.9, ceph to 16.2.10, and ceph-csi to 3.6.2.  Recent rook releases now properly support managing nfs with ceph >= 16.2.7. Also note that rook 1.10 will drop support for ceph 15.x and all clusters need to be migrated to ceph >= 16 in order to update rook in the future. The changes developed on ruka were then copied to the other clusters as a single commit.

Other notable changes are:

- ceph pool quotas set on CRD resources
- most ceph related pods have `resources.limits` set
- the `ingress` resource for the ceph dashboard is now managed by the `rook-ceph-cluster` chart


jhoblitt:
I'm converting this to a draft as I've discovered I need to make changes to the rook-ceph-cluster chart values.

cbarria (approved):
needed upgrade and improvement, i believe the problem started with the quotas exceeding the real capacity in some cases. 👍

cbarria (approved):
changes in ceph values 👍

PR #104, opened 2022-09-09
(Pillan) cloudnativePG deployment

cbarria:
Deployment of cloudnativePG was needed in TTS (tested previously on Ruka Cluster)


jhoblitt:
IT-3771 is the epic. It is probably better to attach this code to the IT-3963 ticket as down the road, people are more likely to find that ticket in a search rather than looking for code on the epic itself.

Also, there is no need to replay the entire commit history from IT-3858.

jhoblitt:
Also, don't forget to rebase on the current master branch.

PR #105, opened 2022-09-09
(pillan) Deployment for cloudnativePG

cbarria:
This is the cloudnativePG deployment needed at TTS and tested on the ruka cluster first.


jhoblitt (approved):
LGTM

PR #109, opened 2022-10-03
Standardize ingress nginx script to match Helm chart

cbarria:
Three clusters still had the old script name.
Now yagan, ruka and pillan have the same name in the script as the rest of the cluster. ('ingress-nginx.sh')


jhoblitt (approved):
LGTM

PR #112, opened 2022-10-21
Upgrade Metallb to v1.13.7 with Helm Chart

cbarria:
Upgraded Metallb to v1.13.7 and change installation from Manifest to Helm Chart, tested and working in Ruka.
Each instance of metallb had their own network segment as follows :

- Chonchon - 139.229.160.201-139.229.160.210
- Gaw - 139.229.135.201-139.229.135.210
- Kueyen - 139.229.144.50-139.229.144.100
- Luan - 139.229.135.150-139.229.135.200
- Lukay - 139.229.160.211-139.229.160.220
- Pillan - 140.252.146.46-140.252.146.59
- Rancher.cp - 139.229.160.24/32
- Rancher.ls - 139.229.160.24/32
- Rancher.tu - 140.252.146.84/32
- Ruka - 139.229.134.150-139.229.134.200g
- Yagan - 139.229.160.150-139.229.160.200
- Yepun - 139.229.160.127-139.229.160.149


chonchon/metallb/metallb.sh
```diff
@@ -1,10 +1,9 @@
...
-
-kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/namespace.yaml
-kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/metallb.yaml
-# On first install only
-kubectl create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
-
-kubectl apply -f ./mlb-l2.yaml
+helm repo add metallb https://metallb.github.io/metallb
+helm --namespace metallb-system \
+    install --create-namespace \
```
jhoblitt, line 6: Please use `update --install` instead of `install` so that the script can be rerun without modification.

jhoblitt (changes requested):
This looks reasonable to me. My only request is to make the install scripts re-runnable.  I did not visually check that the IP pools haven't changed for every cluster.

chonchon/metallb/metallb.sh
cbarria: thats a good idea, i will make the changes to the scripts. About the pools, i just keep the same pools that were configure before, we will change the yagan one for the requierement later on. Thx

jhoblitt (approved):
LGTM.

PR #115, opened 2022-11-03
CloudnativePG Manual Backup improvements

cbarria:
- Adds 60 Days Lifecycle on AWS S3 Storage. (vs Lifecycle on AWS which is paid)
- Adds 5Gi as Ephemeral Storage to avoid running out of space while backing data up.
- Stop logging, bump Helm Chart version to 0.15.0 (which has v1.17.1 of cnpg), disable native backup (commented)

### DIFF
--- a/ruka/cnpg/cnpg.sh
+++ b/ruka/cnpg/cnpg.sh
@@ -5,7 +5,7 @@ set -xe
 # Install cloudnativePG on cluster
 helm repo add cnpg https://cloudnative-pg.github.io/charts
 helm upgrade --install cnpg \
-  --version="0.14.0" \
+  --version="0.15.0" \
   --namespace cnpg-system \
   --create-namespace \
   cnpg/cloudnative-pg \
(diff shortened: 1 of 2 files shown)


ruka/cnpg/cnpg.sh
```diff
@@ -77,28 +77,28 @@ spec:
...
-          name: cnpg-aws-creds
-          key: ACCESS_KEY_ID
-        secretAccessKey:
-          name: cnpg-aws-creds
-          key: ACCESS_SECRET_KEY
-      wal:
-          compression: gzip
-
-    retentionPolicy: "90d"
+#  backup:
```
jhoblitt, line 80: Why is the backup block being commented out instead of deleted?

cbarria: because i don't want to lose the block so i can test later on the normal backups when other version come out, but on the TTS and Summit deployment the backup block will be gone.

jhoblitt: OK. My general preference is to delete dead configuration / code.

jhoblitt (approved):

PR #140, opened 2023-01-25
(manke) manke08, manke09 & manke10 role worker only

cbarria:
in order to avoid problems with etcd, we will use only 7 host on role etcd/controlplane


jhoblitt (approved):
I would prefer that the disabled role labels are deleted rather than commented out as I think it is unnecessary noise but it is acceptable.

PR #148, opened 2023-02-28
(manke) Ceph dashboard ingress standalone

cbarria:
This adds an standalone yaml file for the rook-ceph ingress in case the ArgoCD deployment erases it.


jhoblitt:
@cbarria I'm not following why this is necessary?  SQRE argocd is replacing the ingress controller and not ingress resources.

cbarria:
oh i see, never mind then... i notice the ingress was missing on the namespace that's why it wasn't working, thought was related to Argo.

PR #154, opened 2023-03-10
Konkong k8s Cookbook Operators

cbarria:
This adds the following to the cookbook for the Konkong cluster:
- rke
- metalLB (added lhn)
- nginx Ingress
- cert-manager
- multus on the bridge interface

### DIFF
--- a/konkong/rke/cluster.yml
+++ b/konkong/rke/cluster.yml
@@ -1,0 +1,163 @@
+---
+# If you intened to deploy Kubernetes in an air-gapped environment,
+# please consult the documentation on how to configure custom RKE images.
+nodes:
+- address: konkong01.ls.lsst.org
+  hostname_override: konkong01
+  user: rke
+  role:
+  - controlplane
+  - worker
+  - etcd
+  labels:
+    role: storage-node
+    jupyterlab: ok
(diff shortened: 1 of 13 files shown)


konkong/rke/cluster.yml
```diff
@@ -0,0 +1,163 @@
...
+  role:
+  - controlplane
+  - worker
+  - etcd
+  labels:
+    role: storage-node
+    jupyterlab: ok
+#- address: konkong04.ls.lsst.org
+#  hostname_override: konkong04
+#  user: rke
```
hreinking, line 37: Remember to use only 3 etcd nodes

hreinking (approved):
lgtm

konkong/rke/cluster.yml
cbarria: yes, will be adjusted when all 5 nodes are ready (to only 3etcd, 5 is too overkill for a 5 node cluster)

jhoblitt: There's no reason **not** to use 5 etcd nodes for a 5 node cluster.

jhoblitt (approved):

PR #155, opened 2023-03-14
Metallb 0.13.9 version Bump

cbarria:
Josh found that metallb version 0.13.7 was suddenly stop working after a few hours on the LHN link project, by updating to version 0.13.9, all clusters have been stable.


jhoblitt (approved):
Thank you for doing this.

PR #157, opened 2023-03-30
Kueyen Cookbook Updates for EL8

cbarria:
- RKE was modified for kueyen 01-03
- rook-ceph was modified to work with current hardware


csilva-cl (changes requested):
you mixed the re-provision of the cluster with the testing of cnpg replica, can you please split the PR in two so we can have a separate history for each change ?

csilva-cl (approved):
looks good, thanks.

PR #160, opened 2023-04-04
Update rook-ceph config for TTS

cbarria:
quotas been reach for s3 and the replication pool, plus some config debt on dashboard.

### DIFF
--- a/pillan/rook-ceph/cephblockpool.yaml
+++ b/pillan/rook-ceph/cephblockpool.yaml
@@ -10,4 +10,4 @@ spec:
     size: 3
     requireSafeReplicaSize: true
   quotas:
-    maxSize: 3.4Ti
+    maxSize: 5Ti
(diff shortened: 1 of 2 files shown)


jhoblitt (commented):
I'm not excited about using `ceph-storagepool.yaml`  as a file name. The resource kind is `CephBlockPool` and the name is `replicapool` -- neither of which is reflected in the file name.

cbarria:
was the only filename in the directory without a dash, all others have rook- or ceph- but you are right, the resource is called `CephBlockPool`, changes were reversed. 👍🏻

jhoblitt (approved):

jhoblitt:
Please don't merge this branch while it contains a merge commit.

cbarria:
i manage to notice first! its fixed now and reflects the real quota, will have to set route53 tho

PR #167, opened 2023-04-17
(ruka) Add LHN ingress controller for rook-ceph

cbarria:
This adds a second ingress controller and add the ingress for the s3 on rook as part of the test we did to implement this into chonchon.

### DIFF
--- a/ruka/ingress/lhn/ingress-s3-lhn.yaml
+++ b/ruka/ingress/lhn/ingress-s3-lhn.yaml
@@ -1,0 +1,24 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  annotations:
+    cert-manager.io/cluster-issuer: letsencrypt
+    kubernetes.io/ingress.class: nginx-lhn
+  name: rook-ceph-rgw-ingress-lhn
+  namespace: rook-ceph
+spec:
+  rules:
+  - host: s3-lhn.ruka.dev.lsst.org
+    http:
+      paths:
+      - backend:
+          service:
+            name: rook-ceph-rgw-lfa
+            port:
+              number: 80
+        path: /
+        pathType: Prefix
+  tls:
+  - hosts:
+    - s3-lhn.ruka.dev.lsst.org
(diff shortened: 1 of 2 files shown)


ruka/ingress/lhn/ingress-s3-lhn.yaml
```diff
@@ -0,0 +1,28 @@
...
+            name: rook-ceph-rgw-lfa
+            port:
+              number: 80
+        path: /
+        pathType: Prefix
+  tls:
+  - hosts:
+    - s3-lhn.dev.lsst.org
+    secretName: rook-ceph-rgw-ingress-lhn-tls
+status:
```
jhoblitt, line 25: AFAIK, the status hash can not be written to. Also, it isn't possible to set an IP on an ingress resource as the svc ip is a property of the ingress controller, not the ingress resource.

ruka/ingress/lhn/ingress-s3-lhn.yaml
```diff
@@ -0,0 +1,28 @@
...
+kind: Ingress
+metadata:
+  annotations:
+    cert-manager.io/cluster-issuer: letsencrypt
+    kubernetes.io/ingress.class: nginx-lhn
+  name: rook-ceph-rgw-ingress-lhn
+  namespace: rook-ceph
+spec:
+  rules:
+  - host: s3-lhn.dev.lsst.org
```
jhoblitt, line 11: since this isn't an end-user facing service, how about s3-lhn.ruka.dev.lsst.org?

ruka/ingress/lhn/ingress-s3-lhn.yaml
cbarria: so sorry i took the ingress from somewhere with -o yaml and deleted the garbage, but i missed this one!

ruka/ingress/lhn/ingress-s3-lhn.yaml
cbarria: i had it like that, going to change it back 👍🏻

jhoblitt (approved):
LGTM.

PR #174, opened 2023-06-13
(yagan) add auxtel nfs export

jhoblitt:
Intended to replace the auxtel-archiver.cp.lsst.org:/repo export.


cbarria (approved):
Looks Great 👍🏻  thanks for fixing the script also!

PR #176, opened 2023-06-16
(pillan) add metallb "general" pool; add auxtel nfs export; pin nfs svc IPs

jhoblitt:
The current metallb pool on pillan is full and can't be expanded as the subnet is also full. A new /27 has been allocated to be the new default pool while leaving the existing pool in place.

This is not a complete solution as new routes will also be needed to make this work but can't be applied as this time due to release testing on pillan.


cbarria (approved):
Looks Great 👍🏻

dtapiacl (approved):
LGTM.

PR #181, opened 2023-06-21
(manke) add nfs-comcam export

jhoblitt:
While there isn't currently a "comcam" in the BTS, an nfs export for comcam is being added to stay consistent with TTS.


cbarria (approved):
Looks Great 👍🏻

PR #183, opened 2023-06-23
(* rook clusters) improve rook-ceph.sh; ensure ceph orch is enabled; enable ceph device monitoring;

jhoblitt:
This will renormalize the rook-ceph.sh script across all clusters after the recent per cluster nfs improvements and is intended to prepare for upgrading ceph to 17.2.6.


cbarria (approved):
Looks Great 👍🏻

PR #189, opened 2023-07-18
(manke) add disks and set replicapool quota 15Ti

cbarria:
influxDB needs two deployments at base (manke) in order to restore when data is lost, so 10 Disk 1.92Tb were added and the rook-ceph deployment was adjusted to hold this two instances (5tb ea~)


csilva-cl (approved):
looks good

PR #192, opened 2023-07-24
cert-manager and ingress-nginx version upgrade

cbarria:
its needed to move forward with the current versions of both deployments in preparation to do k8s upgrade. (and because we are going out of support/EOL)


jhoblitt (approved):
LGTM.

PR #193, opened 2023-07-25
(*) normalize CephFilesystem to 4cpu/4Gi & CephNFS to 3cpu/8Gi

jhoblitt:
Based on testing on pillan, the 2Gi CephFilesystem (mds) memory limit on yagan may be insufficient during startup. All rook-ceph clusters are being normalized to use the same resource limits.


cbarria (approved):
Looks Great 👍🏻

PR #196, opened 2023-08-16
(konkong) fix lfa ingress for o11y s3

cbarria:
the ingress for the o11y access had the same name as the original lfa s3 ingress so it got overwritten... also changed the certs issuer to cluster-issuer both certs got signed correctly.


jhoblitt (approved):
Oops. Thanks for cleaning up my mess.

PR #198, opened 2023-08-29
deploy blackbox exporter and snmp exporter

fbegyn:
This uses the prometheus-community helm charts to get blackbox and snmp exporter up and runnning in the cluster. Secrets should be looked into when updating the snmp community strings.


cbarria (changes requested):
can you please add the atomic flag to the helm charts? so it roll back if the installation process if something goes wrong

PR #199, opened 2023-08-29
Ayekan mimir deployment

fbegyn:
Deploys Mimir to the ayekan cluster and configures the in-cluster prometheus to start sending data into Mimir.

### DIFF
--- a/ayekan/prometheus/values.yaml
+++ b/ayekan/prometheus/values.yaml
@@ -1,5 +1,8 @@
 ---
 prometheus:
+  prometheusSpec:
+    remoteWrite:
+      - url: http://ayekan-mimir-gateway.mimir.svc:80/api/v1/push
   ingress:
     enabled: true
     annotations:


ayekan/mimir/mimir.yaml
csilva-cl, line 1: We should make the s3 credentials a env. variable.

csilva-cl (changes requested):

cbarria (changes requested):
can you please change the commit messages to "(ayekan) commit msg" so we can keep the same notation for each cluster?
also now with the Jira account, branches use the ticket name to be associated with the ticket, let me know, i can explain further 👍🏻

ayekan/mimir/mimir.yaml
```diff
@@ -0,0 +1,197 @@
...
+mimir:
+  structuredConfig:
+    common:
+      storage:
+        backend: s3
+        s3:
+          endpoint: "s3.o11y.ls.lsst.org"
+          region: "o11y"
+          access_key_id: "ZE6RTRGFTWG1HV30NMO9"
+          secret_access_key: "VqBygAQpxmTLW48Py5hyC3lpMyZnFP21Kt0mI2xr"
```
jhoblitt, line 11: This is a public repo...

ayekan/mimir/mimir.yaml
```diff
@@ -0,0 +1,197 @@
+# Mimir components
+mimir:
+  structuredConfig:
+    common:
+      storage:
+        backend: s3
```
jhoblitt, line 6: Looking at the `values.yaml` for the mimir chart, it doesn't appear that there is support for using a k8s secret for the s3 credentials. https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/values.yaml

This is sub-optional as it make it difficult for us to use an external secrets operator.  For the moment, we are going to need to resort to interpolating a value from an env var into the yaml.

PR #203, opened 2023-09-05
(chango) Chango cookbook configurations

dtapiacl:


chango/rook-ceph/scripts/rook-ceph-node-cleanup.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -x
+
+DEV1=/dev/sda
+DEV2=/dev/sdb
+SSH_CMD='clush -g ruka'
```
jhoblitt, line 7: I'm guessing this script has never been used. I would suggest simply removing it rather than having an incorrect/unsafe script in the repo.

dtapiacl: Thank you for the feedback, I totally agree. The scripts folder was removed.

jhoblitt (approved):
This looks reasonable to me.

PR #208, opened 2023-10-05
(pillan/manke/yagan) add pg_sphere extension to database

cbarria:
was requested to add support for pg_sphere on the 3 main databases (TTS, BTS & Summit)


jhoblitt (approved):
Superficially, this looks reasonable to me but I'm not really familiar with the chart.

How is the lsstit/cnpgsphere OCI image being built?

cbarria:
i will make the repo with the Dockeimage and setup GHA 👍🏻

PR #209, opened 2023-10-05
(yepun) add cloudnativePG deployment

cbarria:
Keycloak needed a database outside the pixel zone and Cristian asked to be in Yepun

### DIFF
--- a/yepun/cnpg/cnpg.sh
+++ b/yepun/cnpg/cnpg.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+set -xe
+
+# Install cloudnativePG on cluster (chart 0.17.0 installs 1.19.0 cnpg)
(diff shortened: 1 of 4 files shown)


yepun/cnpg/cnpg.sh
```diff
@@ -0,0 +1,70 @@
...
+  username: $(echo -n "postgres" | base64)
+kind: Secret
+metadata:
+  name: cnpg-cluster-superuser
+  namespace: cloudnativepg
+type: kubernetes.io/basic-auth
+END
+
+# Deployment FIRST TIME ONLY
+cat > deploy.yaml << END
```
jhoblitt, line 32: I realize we do this all over the place, but as a general comment, inlining yaml inside a bash script does prevent it from being linted. This is something we should discuss and figure out what we want our standard to be.

jhoblitt (approved):

PR #211, opened 2023-10-24
(ayekan) Configure Prometheus for Mimir remote write and deploy Mimir

fbegyn:
This deploys the Mimir cluster to Ayekan and ensures that the local prometheus writes to it. It also configures the prometheus to only use serviceMonitors from `o11y.eu/monitor=enabled` namespaces.

### DIFF
--- a/ayekan/prometheus/prometheus.sh
+++ b/ayekan/prometheus/prometheus.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 
 set -ex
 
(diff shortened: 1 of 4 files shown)


ayekan/prometheus/prometheus.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
```
jhoblitt, line 1: The space after the `#!` is much less common than no space.

jhoblitt (changes requested):
This looks good other than than the small nitpick about the shebang.

fbegyn:
Fixed the shebang nit

ayekan/prometheus/prometheus.sh
fbegyn: fixed it in the last commit

jhoblitt (approved):
LGTM.

PR #215, opened 2023-11-15
(ayekan) Setup basic exporters to scrape machines

fbegyn:
This should cause the blackbox exporter to start probing all the node-exporter machines.

TODO: configure some SNMP targets as well


cbarria (approved):
Looks Great 👍🏻

PR #216, opened 2023-11-15
Add s3 for rubintv on TTS, BTS and Summit.

cbarria:
as per ticket IT-4841, 3 new buckets are needed for rubinTV, changes are the following:

- standardize the lfa as rgw-lfa.yaml instead of ingress and cephobjectstore.
- added rgw-rubintv.yaml
- changed the rook-ceph.sh script to apply this changes.


chonchon/rook-ceph/s3/rgw-rubintv.yaml
```diff
@@ -0,0 +1,58 @@
...
+      size: 3
+    quotas:
+      maxSize: 10Gi
+  dataPool:
+    failureDomain: host
+    erasureCoded:
+      dataChunks: 2
+      codingChunks: 1
+    quotas:
+      maxSize: 5Ti
```
jhoblitt, line 20: Chonchon does not have the capacity to support 8Ti LFA + 5Ti rubintv.  The quota sizes should be reduced so we aren't overcommiting the storage.

jhoblitt (changes requested):
This generally looks good but I'm concerned about the size of the quotas on chonchon.

jhoblitt (approved):
LGTM. Thank you doing the due diligence on the storage capacity.

PR #220, opened 2023-12-06
(ayekan) enable slack/squadcast connection for alertmanager

fbegyn:
This enabled the slack connection from alertmanager, enabling alert sending to slack. Meanwhile also correctly setting the externalURL of both prometheus and alertmanager (to make the alerts functional).

Only a DNS entry for `alertmanager.ayekan.ls.lsst.org` needs to be configured (as currently the links fail).

If somebody knows a better way around `--set "alertmanager.config.receivers[1].slack_configs[0].api_url=${O11Y_SLACK_API_URL}` I'd be happy to replace the fragile bash.

### DIFF
--- a/ayekan/prometheus/prometheus.sh
+++ b/ayekan/prometheus/prometheus.sh
@@ -9,7 +9,6 @@ helm repo update
 helm upgrade --install \
   kube-prometheus-stack prometheus-community/kube-prometheus-stack \
   --create-namespace --namespace kube-prometheus-stack \
-  --set "alertmanager.config.receivers[1].slack_configs[0].api_url=${O11Y_SLACK_API_URL}" \
   -f ./values.yaml \
   --atomic
 
--- a/ayekan/prometheus/prometheus.sh
+++ b/ayekan/prometheus/prometheus.sh
@@ -9,6 +9,7 @@ helm repo update
 helm upgrade --install \
   kube-prometheus-stack prometheus-community/kube-prometheus-stack \
   --create-namespace --namespace kube-prometheus-stack \
+  --set alertmanager.config.receivers.[1].slack_configs.[1].api_url ${O11Y_SLACK_API_URL}
   -f ./values.yaml \
   --atomic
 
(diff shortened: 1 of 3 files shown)


ayekan/prometheus/prometheus.sh
```diff
@@ -9,6 +9,7 @@ helm repo update
 helm upgrade --install \
   kube-prometheus-stack prometheus-community/kube-prometheus-stack \
   --create-namespace --namespace kube-prometheus-stack \
+  --set "alertmanager.config.receivers[1].slack_configs[0].api_url=${O11Y_SLACK_API_URL}" \
```
cbarria, line 12: why not use this inside your values.yaml?
`alertmanager:
  config:
    global:
      resolve_timeout: 5m
      # The API URL to use for Slack notifications.
      [ slack_api_url: <secret> ]
      [ slack_api_url_file: <filepath> ]`
     [](https://prometheus.io/docs/alerting/latest/configuration/#configuration-file)

fbegyn: The kubernetes secret way? (I'm guessing - There are 2 methods in the comment above 'that way' doesn't really reference 1 specific)

cbarria: yo asked for a replacement for this line: 
`--set "alertmanager.config.receivers[1].slack_configs[0].api_url=${O11Y_SLACK_API_URL}"`
I offer 2 solutions, I guess any of those is better than having both... a 'values.yaml' and a -set

jhoblitt: I think that mixing `--set` with a values file is going to lead to confusion in the future.  I would prefer that it is one or the other but never both.

fbegyn: As discussed, I've moved the Slack secret into a k8s secret that gets mounted into the container to make the bash scripts cleaner. I think that resolves a discussion point for this PR.

jhoblitt (approved):
LGTM.

PR #225, opened 2024-01-16
(ayekan) add kyverno

jhoblitt:


ayekan/kyverno/kyverno.sh
```diff
@@ -0,0 +1,69 @@
...
+  tag: 'v1.8.5'
+replicaCount: 3
+resources:
+  limits:
+    cpu: '100m'
+    memory: '1Gi'
+initResources:
+  limits:
+    cpu: '100m'
+    memory: '1Gi'
```
fbegyn, line 66:
```suggestion
    memory: '1Gi'
admissionController:
  metricsService:
    create: true
backgroundController:
  metricsService:
    create: true
cleanupController:
  metricsService:
    create: true
reportsController:
  metricsService:
    create: true
```

For the Kyverno metrics, [some modifications](https://kyverno.io/docs/monitoring/) are needed it seems. Creating the metrics service for all the compenents by enabling these variables should do the trick.

ayekan/kyverno/policies/require_pod_requests_limits.yaml
```diff
@@ -0,0 +1,38 @@
...
+kind: ClusterPolicy
+metadata:
+  name: require-requests-limits
+  annotations:
+    policies.kyverno.io/title: Require Limits and Requests
+    policies.kyverno.io/category: Best Practices
+    policies.kyverno.io/severity: medium
+    policies.kyverno.io/subject: Pod
+    policies.kyverno.io/description: >-
+      As application workloads share cluster resources, it is important to limit resources
```
fbegyn, line 12: Does Kyverno actively enforce the policies? Or only when a new config is applied to the cluster through admissioncontrollers?

This one could cause some havoc if applied retro-active (although I agree with the permise).

fbegyn (changes requested):
Please enable the `metricsService` for each component. Otherwise all things look in order.

ayekan/kyverno/policies/require_pod_requests_limits.yaml
jhoblitt: All of the policies are in audit mode, which means no enforcement.

fbegyn: LGTM then

fbegyn (approved):
LGTM

PR #230, opened 2024-01-17
(ayekan) add puppetdb exporter to ayekan and connect to dev instance

fbegyn:
Deploys the puppetdb exporter instance to ayekan cluster and connects to the dev instance.


ayekan/exporters/puppetdb-exporter-deployment.yaml
```diff
@@ -0,0 +1,128 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    o11y.eu/exporter: "kafka"
```
jhoblitt, line 6: How is this related to kafka?

fbegyn: copy/pate issues and too many open branches I think :sweat_smile:

ayekan/exporters/puppetdb-exporter-deployment.yaml
```diff
@@ -0,0 +1,206 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
jhoblitt, line 4: A `namespace` should be set on all resources going forward as I'm planning to start migrating deployments into fleet starting this week but I don't want to hold this up any further so I'll mark the PR approved.

jhoblitt (approved):

PR #232, opened 2024-01-25
(*) add external-secrets operator

jhoblitt:
Note that installing external-secrets (and charts which it depends upon only on the rancher.* clusters) is a special case  / bootstraping and will continue to require the use of a shell script, env vars, and/or templating until secret inject is handled by a CI/CD tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,5 @@
 k8s cookbook
 ============
+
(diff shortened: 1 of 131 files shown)


cbarria (approved):
Looks Great 👍🏻

csilva-cl (changes requested):
it looks like fetch-credentials needs to be executed before the operator is added, could be good to add a small readme with those instructions.

also the 1pass CLI must be installed.

csilva-cl (approved):
Thanks for the changes.

Looks good

PR #242, opened 2024-01-31
(pillan) convert keycloak to use external-secrets

jhoblitt:


csilva-cl (commented):
Looks good, but I can't find the secrets in 1password. Is that expected ?

jhoblitt:
> Looks good, but I can't find the secrets in 1password. Is that expected ?

The secrets have been moved to the `tu` vault.

PR #243, opened 2024-01-31
(rancher.*) bump rancher to 2.7.9

jhoblitt:
See https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/security-advisories-and-cves


cbarria (approved):
Looks Great 👍🏻

PR #247, opened 2024-02-05
(ayekan) fix prometheus alerts for missing k8s components

fbegyn:
Due to Rancher, some components are missing from their standard places in kubernetes and this triggers unactionable alerts from the prometheus stack.

### DIFF
--- a/ayekan/prometheus/values.yaml
+++ b/ayekan/prometheus/values.yaml
@@ -219,3 +219,11 @@ grafana:
           jsonData:
             handleGrafanaManagedAlerts: false
             implementation: prometheus
+
+# these components are not present in the Rancher clusters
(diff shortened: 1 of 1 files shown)


ayekan/prometheus/values.yaml
```diff
@@ -219,3 +219,11 @@ grafana:
           jsonData:
             handleGrafanaManagedAlerts: false
             implementation: prometheus
+
+# these components are not present in the Rancher clusters
+kubeScheduler:
```
jhoblitt, line 224: I'm not sure I'm following... All 3 of these processes are running on the ayekan nodes.

jhoblitt (approved):

PR #258, opened 2024-03-04
(pillan) initial fleet migration for pillan

jhoblitt:
Note that rancher.tu uses `c-st7p2` for pillan.  It appears that pillan needs to be deleted from rancher and re-imported to fix this.  This old rancher naming convention seems to have already been fixed for all other clusters.

The following services are migrated to be being deployed via fleet:

- cert-manager
- external-secrets
- kyverno
- metallb
- multus
- rook-ceph


cbarria (approved):
LGTM

PR #264, opened 2024-03-07
(fleet/kube-prometheus-stack) bump kube-prometheus-stack to v56.19.0

jhoblitt:
and add the `o11y.eu/monitor` label to the `kube-prometheus-stack` ns.

This will harmonize the entire fleet with the version deployed on ayekan.


fbegyn (approved):
Not fully grokking the `jsonPointers`, but looks good to me.

jhoblitt:
`jsonPointers` is telling fleet to ignore changes to those parts of the resource when determining if the resource is in sync.  This is necessary as many operators remove configuration that has default values and/or insert values.

PR #267, opened 2024-03-07
(pillan) add o11y rgw

jhoblitt:
Provision a new rgw instance accessible as s3.o11y.tu.lsst.org to sort metrics and logs for the TTS.


fbegyn (approved):
LGTM

PR #268, opened 2024-03-07
(rancher.*) fix missing onepassword chart values

jhoblitt:
These values.yaml files were unintentionally excluded by the root .gitignore.


cbarria (approved):
Looks Great 👍🏻

PR #276, opened 2024-03-15
(ruka) add mimir

jhoblitt:


cbarria (approved):
Looks Great 👍🏻

fbegyn (approved):
I'm not a total fan of the current structure for the prometheus-operator deployment. But that can be handled in a future PR. For deployment this makes sense and looks good.

PR #277, opened 2024-03-20
Prometheus deployment rework

fbegyn:
Currently, only a single prometheus operator is allowed in a cluster (or very strict control over namespaces is needed). This branch makes an effort to split up the deployment of the prometheus operator and a prometheus stack for the o11y metrics project. It also rethinks the label used for monitoring discovery.

### DIFF
--- a/ayekan/exporters/blackbox-exporter/helm-values.yaml
+++ b/ayekan/exporters/blackbox-exporter/helm-values.yaml
@@ -14,10 +14,10 @@ serviceMonitor:
   selfMonitor:
     enabled: true
     labels:
-      monitor: enabled
+      lsst.io/monitor: "true"
   defaults:
     labels:
-      monitor: enabled
+      lsst.io/monitor: "true"
   targets:
     - name: googlehttpcheck
       url: https://google.com
@@ -25,8 +25,14 @@ serviceMonitor:
       interval: 60s
       module: http_2xx
 
+# TODO: Error: UPGRADE FAILED: release blackbox-exporter failed, and has been
+# rolled back due to atomic being set: failed to create resource: admission
+# webhook "prometheusrulemutate.monitoring.coreos.com" denied the request: Cannot
+# unmarshal rules from spec
 prometheusRule:
   enabled: false
+  labels:
+    lsst.io/discover: "true"
 
 extraConfigmapMounts: []
 extraSecretMounts: []
--- a/fleet/lib/ingress-nginx/fleet.yaml
+++ b/fleet/lib/ingress-nginx/fleet.yaml
@@ -18,7 +18,11 @@ helm:
           enabled: true
           additionalLabels:
             lsst.io/monitor: "true"
-        # TODO: handle error: cannot unmarshal rule from spec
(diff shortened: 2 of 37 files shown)


fbegyn:
Reworked and rebased on the current master. The goal here is to re-align all the prometheus labels and instances with the work done the past few weeks. A lot of serviceMonitors are currently not managed or scraped because of missing labels and configuration.

ayekan/exporters/blackbox-exporter/helm-values.yaml
```diff
@@ -14,19 +14,25 @@ serviceMonitor:
...
     - name: googlehttpcheck
       url: https://google.com
       hostname: google.com
       interval: 60s
       module: http_2xx
 
+# TODO: Error: UPGRADE FAILED: release blackbox-exporter failed, and has been
+# rolled back due to atomic being set: failed to create resource: admission
+# webhook "prometheusrulemutate.monitoring.coreos.com" denied the request: Cannot
+# unmarshal rules from spec
```
jhoblitt, line 31: This error shouldn't come up under fleet as we switched from atomic to fleets equivalent of `-wait-for-jobs`.

fleet/lib/ingress-nginx/fleet.yaml
```diff
@@ -10,6 +12,17 @@ helm:
   values:
     controller:
       kind: DaemonSet
+      metrics:
+        enabled: true
+        serviceMonitor:
+          enabled: true
+          additionalLabels:
+            lsst.io/monitor: "true"
+        # TODO: handle error: cannot unmarshal rule from spec
```
jhoblitt, line 21: It sounds like this would break the deployment in other clusters?  If so, this should either be changed to a targetCustomization or split out into its own PR.

jhoblitt (changes requested):
Overall, this looks good. My only concern is with the `ingress-nginx` deployment.  Those changes should either be restricted to ayekan or split out into an separate PR until deployment error is fixed.

ayekan/exporters/blackbox-exporter/helm-values.yaml
fbegyn: I think the issue is that the `PrometheusRule` objects inside the blackbox-exporter is breaking on the version of the Prometheus operator. The error code is copy/pasted from the manual debugging session, so the `atomic` output might be incorrect. I'll correct the TODO so it correctly reflects the problem (since there rules still break when deploying with rancher).

fleet/lib/ingress-nginx/fleet.yaml
fbegyn: The TODO applies to the lines lower. The `prometheusRules` included on the chart can't be unmarshalled by the current Prometheus operator format. I'll adjust the TODO to be more clear and correct.

fbegyn:
@jhoblitt I think this is ready for a merge now. Been rebased and all checks pass. Once this is in master all clusters can get synced up.

jhoblitt (approved):

PR #301, opened 2024-04-01
(fleet/multus) mv network-attachment-definition-* -> net-attach-def-*

jhoblitt:
To be consist with the file naming convention of `<kind name>-<resource name>` where the kind "shortname" is preferred, if one is set.


cbarria (approved):
Looks Great! 👍🏻

PR #305, opened 2024-04-04
(pillan) mv cnpg to fleet

cbarria:
Attempt to migrate cloudnativePG into Fleet for the cluster in pillan.


fleet/lib/cnpg-cluster/fleet.yaml
```diff
@@ -17,3 +17,42 @@ dependsOn:
...
+    yaml:
+      overlays:
+        - ls
+  - name: cp
+    clusterSelector:
+      matchLabels:
+        site: cp
+    yaml:
+      overlays:
+        - cp
```
jhoblitt, line 58: This patch doesn't exist.

fleet/lib/cnpg-cluster/fleet.yaml
```diff
@@ -17,3 +17,42 @@ dependsOn:
...
+    yaml:
+      overlays:
+        - tu
+  - name: ls
+    clusterSelector:
+      matchLabels:
+        site: ls
+    yaml:
+      overlays:
+        - ls
```
jhoblitt, line 51: This path doesn't exist.

fleet/lib/cnpg-cluster/fleet.yaml
```diff
@@ -17,3 +17,42 @@ dependsOn:
...
+    clusterName: pillan
+    yaml:
+      overlays:
+        - pillan
+  - name: ruka
+    clusterName: ruka
+    yaml:
+      overlays:
+        - ruka
+  - name: dev
```
jhoblitt, line 31: `targetCustomizations` are applied as the first match starting from the start of the array.  A cluster will only match once.  As an example, `ruka` would match the `ruka` customization but never the `dev` customization as `ruka` is in the array before `dev`.

It appears that you are trying to merge the `pillan` + `tu` and `ruka` + `dev` customizations, which won't work.  A means of handling this would be to have multiple overlay dirs for each of the `pillan` and `ruka` targets.  An example of doing this (except with helm values) is https://github.com/lsst-it/k8s-cookbook/blob/master/fleet/lib/kube-prometheus-stack/fleet.yaml#L44-L56.

jhoblitt (changes requested):
Please remove the files under `pillan/cnpg` that will be unused after the fleet migration.

fleet/lib/cnpg-cluster/fleet.yaml
cbarria: path is gone 👍🏻

fleet/lib/cnpg-cluster/fleet.yaml
cbarria: path removed also

fleet/lib/cnpg-cluster/fleet.yaml
cbarria: since the secrets are specific for each cluster I moved the secrets inside their respective cluster and modified the fleet.yaml to only reflect pillan and ruka.

jhoblitt (approved):
LGTM.

fleet/lib/cnpg-cluster/overlays/pillan/cluster-cnpg-cluster.yaml
```diff
@@ -0,0 +1,43 @@
+---
+# Cluster Definition
+apiVersion: postgresql.cnpg.io/v1
+kind: Cluster
+metadata:
+  name: cnpg-cluster
+spec:
+  imageName: docker.io/lsstit/cnpgsphere:14.5
+  imagePullPolicy: IfNotPresent
```
jhoblitt, line 9: I'm assuming this is intentionally different from the value on ruka?

cbarria: yes, in ruka is set to always pull for testing purposes 👍🏻

PR #311, opened 2024-04-10
(yagan & yepun) initial fleet migration

jhoblitt:
Note that this refactors the metallb config on yepun but leaves yagan alone for the time being.


cbarria (approved):
Looks Great 👍🏻

PR #314, opened 2024-04-10
(cnpg) manke cnpg to fleet

cbarria:
cnpg on manke moved to fleet


jhoblitt (commented):
Could we delete the `manke/cnpg` dir as part of this PR?

jhoblitt (approved):
LGTM.

PR #315, opened 2024-04-10
(fleet/multus) increase request to 64Mi; limit to 128Mi

jhoblitt:
This attempts to resolve the kube-mutlus-ds pods being killed by the oom.


cbarria (approved):
Thanks, Looks Great 👍🏻

PR #317, opened 2024-04-11
(fleet/grafana-dashboards) grafana dashboard GitOps

fbegyn:
Adding a dashboard JSON to the dashboards folder in this directory, will automagically deploy them into the `lsst-grafana-dashboard` namespace in the cluster. If `kube-prometheus-stack` is deployed, it can be configured to pick up this namespace for monitoring:

TODO: decision - maybe install into `kube-prometheus-stack` by default?

Numeric subdirectories in the `/dashboards` folder can be used to assign an organisation ID to the dashboards being uploaded. Dashboards under `/dashboards` are automatically assigned `1`. For examples dashboards under `/dashboards/3` will have organisation ID `3` and under `/dashboards/45` will have organisation ID `45`.


jhoblitt (commented):
Do you know what, if any, significance the numbers postfix on the filename grafana generates when exporting a dashboard mean?  It too many digits to be a count of seconds since the unix epoch.

fbegyn:
@jhoblitt it needs some cleanup, it's just a UNIX timestamp from when I orginally downloaded the dashbaords from the Grafana instance. For example `1706888763471` == `Fri Feb 02 2024 15:46:03 GMT+0000`

jhoblitt:
> `1706888763471`

OK. That would have to include milliseconds then. E.g.

```
$ date --date=@1706888763
Fri Feb  2 08:46:03 AM MST 2024
```

I'm in principal OK with this organization but I think it might be a good idea to strip the timestamp from the files so there will be useful diffs in a PR.

jhoblitt (approved):

PR #320, opened 2024-04-11
(fleet/cnpg-system) increase mem to 192Mi/384Mi

jhoblitt:
This is to resolve the oom killing the cnpg operator.


cbarria (approved):
Looks Great, Thanks!

PR #323, opened 2024-04-12
(dev/*) bump k8s to v1.27.12-rancher1-1

jhoblitt:
The number of allowed pods per node is increased from the default of 110 to 250.

Depends on https://github.com/lsst-it/lsst-control/pull/1127


cbarria (approved):
Looks Great 👍🏻

PR #324, opened 2024-04-15
(fleet/cnpg-cluster) remove yepun & bump v0.5

cbarria:
- Yepun database was empty, Keycloak DB is on each k8s cluster
- There was a missing / on a path when deleting copies over 60 days.


jhoblitt (approved):
There is clearly a lot of duplication between the 4 remaining overlays that should be factored out.  I'm OK with merging this PR as is but the configuration would be DRY'd before making further changes.

PR #326, opened 2024-04-17
(fleet/cnpg-cluster) rm external-secret-cnpg-aws-creds -- unused

jhoblitt:


cbarria (approved):
silly me, now is not needed, thanks!

PR #328, opened 2024-04-18
(fleet/rook-ceph-cluster) enable modern ceph rbd features + discard flag

jhoblitt:
Note that this change will require manual deletion of the `rook-ceph-block` sc.


cbarria (approved):
Looks Great 👍🏻

PR #331, opened 2024-04-24
(fleet/blackbox-exporter) fix ing host & tls secret name

jhoblitt:
The existing cert/secret name is tls-snmp-exporter-ingress which is the same as being used by our snmp-exporter deployment.  This is clearly a typo.


cbarria (approved):
Looks Great 👍🏻

PR #337, opened 2024-04-25
(fleet/blackbox-exporter) add BlackboxExporterModuleUnknown alert

jhoblitt:


cbarria (approved):
Looks Great 👍🏻

fbegyn (approved):
LGTM, but we might wanna allign all the labels before merging these requests. Alertwise and query wise these seem alright.

Generally we use `summary` annotations to create a description that is global without any template variables in it, while `description` annotations will contain specific information about the alert with template variables in it. This allows for a large group of alerts to be grouped together an announced by the summary annotation.

PR #347, opened 2024-05-03
(fleet/prometheus-stack) set the correct key for the grafana keycloak

fbegyn:
As the summary mentions, this was incorrectly set in the 1password. Temporarry adding the correct key, but it should be corrected.

### DIFF
--- a/fleet/lib/kube-prometheus-stack-pre/externalsecret-grafana-keycloak-credentials.yaml
+++ b/fleet/lib/kube-prometheus-stack-pre/externalsecret-grafana-keycloak-credentials.yaml
@@ -18,7 +18,7 @@ spec:
   - remoteRef:
       key: grafana-keycloak-credentials
       property: hostname
-    secretKey: keycloak_url
+    secretKey: url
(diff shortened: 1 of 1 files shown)


fleet/lib/kube-prometheus-stack-pre/externalsecret-grafana-keycloak-credentials.yaml
```diff
@@ -18,7 +18,7 @@ spec:
   - remoteRef:
       key: grafana-keycloak-credentials
       property: hostname
-    secretKey: keycloak_url
+    secretKey: url
```
jhoblitt, line 21: I looked but I can't figure it out. Where is this used?

jhoblitt: I think we should either update `    auth.generic_oauth` to use this value or remove it from the secret.

fbegyn:
Closed in favor of https://github.com/lsst-it/k8s-cookbook/pull/492

PR #348, opened 2024-05-03
(konkong) increase o11y rgw quota to 1.05TiB

jhoblitt:
To allow the pool to function normally as usage is reduced.


csilva-cl (approved):
lgtm

PR #349, opened 2024-05-06
(rancher.dev) initial fleet migration

jhoblitt:


fleet/lib/onepassword-connect/values.yaml
csilva-cl, line 1: is this going to conflict with deployments in other rancher clusters?

jhoblitt: Yes. It needs to be templated.  I'll do that now instead of waiting for additional clusters to be added.

csilva-cl (approved):
lgtm, thank you

PR #377, opened 2024-05-07
(fleet/rook-ceph) PoC for radosgw deployment

fbegyn:
This draft is intended to work out the radosgw usage exporter deployment into the cluster.

source: https://github.com/blemmenes/radosgw_usage_exporter?tab=readme-ov-file

Realistically we probably want this deployed for all the Object stores deployed. This would require either hooking into the `cephObjectStore` value of the `rook-ceph-cluster` or an idea to brainstorm this into the `rook-ceph-conf` layout would be better.

Currently things are hardcoded, while realisitcally speaking based on the `CephObjectStore` name we could template all the other parts of the layout.

### DIFF
--- a/fleet/lib/rook-ceph-conf/charts/ayekan/Chart.yaml
+++ b/fleet/lib/rook-ceph-conf/charts/ayekan/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v2
+apiVersion: v3
 name: ayekan
 description: A Helm chart for Kubernetes
 type: application
--- a/fleet/lib/rook-ceph-conf/charts/ayekan/Chart.yaml
+++ b/fleet/lib/rook-ceph-conf/charts/ayekan/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v3
+apiVersion: v2
 name: ayekan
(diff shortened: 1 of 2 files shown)


fbegyn:
Before I move forward I'm curious if @jhoblitt has some easier insight as to how to get this properly hooked into the current `rook-ceph-conf` sturcture. For the test deployment I can confirm the hardcoded files work.

fbegyn:
Ideally this can be deployed to any namespace, but I'm not fully convinced the rook-ceph operator works in all namespaces.

jhoblitt:
I am not opposed to this if we need the extra data. However, it would only solve the specific case for rgw pool usage (we currently have pools for rbd, cephfs, and rgw) and we need to solve the general case of alerting on any ceph pool that has a quota set and is approaching the limit.  It would also be nice have alert on an individual osds filling up but alerts for pools is critical.  I believe that https://github.com/lsst-it/k8s-cookbook/pull/378 will fix ceph metric scraping and that the built in metrics do expose pool usage & quota.

fbegyn:
Looks like #378 solves the data missing. Closing this in favor of it.

PR #378, opened 2024-05-07
(fleet/rook-ceph) bump chart(s) to 1.14.3 + fix servicemonitor

jhoblitt:


fbegyn (commented):
LGTM. Let's see if the rollout of this handles the metrics fetched in https://github.com/lsst-it/k8s-cookbook/pull/377 . If so we can deprecated that PR, otherwise both might be required.

fbegyn (approved):

PR #392, opened 2024-05-13
(fleet/mimir) set a default metric retention period of 2 years

jhoblitt:
This change has been confirmed as working by inspecting the compactor pod logs:

    ts=2024-05-13T19:03:05.618115801Z caller=blocks_cleaner.go:629 level=info component=cleaner run_id=1715626985 task=clean_up_users user=anonymous msg="marked blocks for deletion" num_blocks=0 retention=17520h0m0s

### DIFF
--- a/fleet/lib/mimir/values.yaml
+++ b/fleet/lib/mimir/values.yaml
@@ -44,6 +44,7 @@ mimir:
       ingestion_burst_size: 1000000
       max_global_series_per_user: 0
       max_global_series_per_metric: 0
+      compactor_blocks_retention_period: 2y
 
 compactor:
   persistentVolume:


fleet/lib/mimir/values.yaml
```diff
@@ -46,6 +46,8 @@ mimir:
       max_global_series_per_metric: 0
```
fbegyn, line 47: It's possible to add `compactor_blocks_retention_period: 2y` here. This has the added benefit of being included in the config hash, which can be used to monitor config drift in a cluster.

fbegyn (commented):
While the `extraArgs` is certainly an option, it's better put under the `structuredConfig` key.

PR #407, opened 2024-05-16
(fleet/kube-prometheus-stack) add externalLabels to all clusters

jhoblitt:
The existing `site` external label would conflict with the labels being set on node-exporter metrics discovered via puppetdb, which is why it has been renamed to `prom_site`.


fbegyn (approved):
LGTM

PR #412, opened 2024-05-17
(konkong) add latiss and lsstcam s3

cbarria:
Four S3 Resources were requested for the BTS
-  rubinobs-raw-latiss (1 TB) - latiss-raw.ls.lsst.org
-  rubinobs-butler-latiss (1 TB) - latiss-butler.ls.lsst.org
-  rubinobs-raw-lsstcam (6 TB) - lsstcam-raw.ls.lsst.org
-  rubinobs-butler-lsstcam (34 TB) - lsstcam-butler.ls.lsst.org


jhoblitt (approved):
This will increase memory usage on konkong by 4GiB x 3 x 4 = ~48GiB.  There is enough unreserved memory for this to work. I'm OK with proceeding with this approach but I suspect at some point we may want to use bucket quotas and a fewer number of rgw instances.

PR #418, opened 2024-05-20
(fleet/kube-prometheus-pre) split alertmanager creds for ayekan & prod

jhoblitt:
This is to use different 1pass item names for prod while leaving the existing ayekan config in place.


cbarria (approved):
Looks Great 👍🏻

PR #419, opened 2024-05-20
(site/ls) refactor 1pass: <cluster> -> k8s-<site> -> k8s-common

jhoblitt:
Cluster level vaults have only been created for:

- antu
- manke

The vaults should be created for other clusters as they are needed.


cbarria (approved):
Looks Great 👍🏻

PR #420, opened 2024-05-20
(cp*) bump v1.27.12

cbarria:
bump k8s version on all cp clusters to v1.27.12

### DIFF
--- a/lukay/rke/cluster.yml
+++ b/lukay/rke/cluster.yml
@@ -39,110 +39,13 @@ nodes:
   labels:
     role: storage-node
 services:
-  etcd:
-    image: ""
-    extra_args:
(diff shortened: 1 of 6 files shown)


jhoblitt (changes requested):
The rancher.cp rancher `values.yaml` should be updated to remove the psp config (as it isn't relevant for k8s > 1.25).

The PR title also needs to be updated.

lukay/rke/cluster.yml
jhoblitt, line 1: The commit message for lukay is capitalized when it shouldn't be.

cbarria: fixed 👍🏻

jhoblitt (approved):
LGTM.

PR #422, opened 2024-05-20
(rancher.cp) into fleet

cbarria:
Moving rancher.cp to fleet! 👍🏻


jhoblitt (commented):
It appears that velero is deployed on rancher.cp and we haven't rolled out a fleetized deployment for it.  Maybe we should leave the rancher.cp/velero dir in place until that is done?

```bash
 ~ $ k get ns velero
NAME     STATUS   AGE
velero   Active   707d
```

jhoblitt (approved):
LGTM now.

PR #425, opened 2024-05-22
(chonchon) reduce cephnfs & cephobjectstore cpu requests

jhoblitt:
The cluster has Insufficient cpu to schedule pods.


cbarria (approved):
Looks Great 👍🏻

PR #431, opened 2024-05-28
(antu) add kube-prometheus-stack

jhoblitt:
This also enables shipping prom metrics from dev, ls, & cp k8s clusters to antu & ruka.


dtapiacl (approved):
LGTM.

PR #433, opened 2024-05-29
(fleet/kube-prometheus-stack) enable scrape labels on "client" clusters

jhoblitt:
At present, only manke is scraping ceph metrics and sending them to ayekan (it isn't clear why that is working). The intent was for all dev, cp, & ls clusters to be scraping and forwarding metrics to mimir on ayekan.


cbarria (approved):
Looks Great 👍🏻

csilva-cl (approved):
looks good, thank you

PR #435, opened 2024-05-29
(fleet/external-secrets-conf) generate clustersecretstore

jhoblitt:
Instead of having a static config per site and/or cluster.


cbarria (approved):
Looks Great 👍🏻  thanks so much for fixing this

PR #442, opened 2024-05-30
(template/external-secrets) create external-secrets ns

jhoblitt:
If the exteranl-secrets ns doesn't already exist, create it.


cbarria (approved):
Looks Great 👍🏻

PR #447, opened 2024-05-30
(fleet/*) rm atomic flag; default is false

jhoblitt:
And it seems that having atomic set may (???) cause fleet to forever retry upgrading a release.


dtapiacl (approved):
LGTM

PR #448, opened 2024-05-30
(fleet/mimir) increase distributor mem to 12Gi

jhoblitt:
To resolve distributor pods being OOMKilled on ruka:

    mimir-distributed-distributor-7767d66476-cdh6f         0/1     OOMKilled     1 (61s ago)     2m4s
    mimir-distributed-distributor-7767d66476-cgf4b         0/1     OOMKilled     1 (72s ago)     2m18s

Increasinig the limit to 8Gi was tested and pods were still getting OOMKilled.

### DIFF
--- a/fleet/lib/mimir/values.yaml
+++ b/fleet/lib/mimir/values.yaml
@@ -62,10 +62,10 @@ distributor:
   replicas: 2
   resources:
     limits:
-      memory: 5Gi
+      memory: 12Gi
     requests:
       cpu: 1
-      memory: 5Gi
+      memory: 12Gi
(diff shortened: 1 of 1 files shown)


fbegyn (approved):
Depending on the data push happening when the cluster spins up, this can be scaled up/down. But this looks good and will be able to handle big pushes.

In an ideal world we world probably have some pod scalers taking care of the resources and we can configure "nominal" working resources here. This would scale up the pod when big pushes happen (on startup of a new cluster for example), but settle back down.

For now LGTM as long as the cluster have these resources available.

PR #455, opened 2024-06-03
(.gitignore) allow values.yaml to be committed

jhoblitt:
This was useful when secrets were being inserted into helm values.yaml files but it is now generally a source of errors with fleet/helm values.yaml accidentally not being committed to the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 creds.sh
 secret.yaml
-values.yaml
 template.yaml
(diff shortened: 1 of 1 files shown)


jhoblitt:
I think yamllint is upset about `.yaml` in the branch name... we should probably exclude `.git`...

jhoblitt:
As there has been no response, I'm going to merge this.

PR #458, opened 2024-06-04
(fleet/snmp-exporter-pre) update snmp-raritian-pdu.json from route53

jhoblitt:
All records in the ls.lsst.org & cp.lsst.org zones which match /pdu/ have been scraped into this update.


hstockeb:
looks good to me

hstockeb (approved):
Looks good to me

PR #465, opened 2024-06-05
(fleet/blackbox-exporter) set cpu resources

jhoblitt:


fleet/lib/blackbox-exporter/values.yaml
```diff
@@ -6,8 +6,10 @@ replicas: 1
 
 resources:
   limits:
+    cpu: "2"
```
fbegyn, line 9: Setting CPU limits isn't really best practice to my knowledge. It can lead to massive throttling occuring and the Linux scheduler punishes the programs extensively.

jhoblitt: If it is a real issue then the request needs to be increased and there are prometheusrules that will warn when a pod is getting throttled.  We are about to throw the switch and enforce that all pods have limits set via kyverno do to real problems we've seen on our systems.  There's a lot written about this. E.g. https://dnastacio.medium.com/why-you-should-keep-using-cpu-limits-on-kubernetes-60c4e50dfc61

fbegyn: I'm not completely opposed to it, but it's a setting that will need some follow up I think. The only reason I personally would use it is for the garanteed qos class. So where requests = limits.

fbegyn (approved):
Keeping in mind we should track the behaviour of this container, LGTM

PR #466, opened 2024-06-05
(fleet/mimir) set ingress & gateway set buffer sizes to 10m

jhoblitt:
Mimir gateway acts as a shadow ingress for mimir components.  Lets keep the ingress resource and gateway buffer sizes in sync.


fbegyn (approved):
LGTM, this makes sense since the massive payloads Prometheus can push sometimes.

PR #470, opened 2024-06-05
(yamllint) only allow quoted string when required

jhoblitt:
Requires https://github.com/bewuethr/yamllint-action/pull/24


cbarria (approved):
Looks Great 👍🏻

PR #473, opened 2024-06-06
(fleet/mimir) disable zoneAwareReplication

jhoblitt:
We don't currently implement ceph availability zones and have no plans to do so.  The zoneAwareReplication functionality in the mimir chart is largely cosmetic in that it will create multiple sts, zone per configured zone, instead of a single sts with multiple instances.  However, when scaling the number of replicas of a service up or down, the full set of zones needs to be manual enumerated and it seems like the minimum number is 3 and we would like to be able to use fewer instances on dev clusters.

### DIFF
--- a/fleet/lib/mimir/values.yaml
+++ b/fleet/lib/mimir/values.yaml
@@ -98,7 +98,7 @@ ingester:
           topologyKey: kubernetes.io/hostname
 
   zoneAwareReplication:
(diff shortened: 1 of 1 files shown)


fbegyn (approved):
LGTM

jhoblitt:
I'm reverting this for the moment as there are problems on ayekan and antu that I don't understand.

PR #474, opened 2024-06-06
(fleet/mimir) change ingestor replication_factor to 2

jhoblitt:
When running 3 ingestor instances and requiring 3 replicas any distriburance to an ingestor pod causes metrics to stop being accepted. This includes rollouts, drains, and node failure.  In order to support operators the number of replicas should always be at least 1 less than the number of ingestor instances.

Changing from 3 to 2 replicas also reduces the ammount of write amplication.

### DIFF
--- a/fleet/lib/mimir/values.yaml
+++ b/fleet/lib/mimir/values.yaml
@@ -48,6 +48,9 @@ mimir:
       compactor_blocks_retention_period: 2y
     compactor:
       deletion_delay: 12h
+    ingester:
+      ring:
+        replication_factor: 2
 
 compactor:
   persistentVolume:


fbegyn (approved):
For some reason, the replication factor documentation never really groks for me. At another site we had the same issue with Cortex (and I believe most of the logic is still shared for this code path).

> With the default replication factor of 3, this means 2 out of 3 writes to ingesters must succeed. If the Mimir cluster loses a minority of ingesters, the in-memory series samples held by the lost ingesters are available in at least one other ingester, meaning no time series samples are lost.

source: https://grafana.com/docs/mimir/latest/references/architecture/components/ingester/

This would lead me to think that a rolling update of the ingesters would be fine here when we have 3 instance running. But seemingly it isn't so the change makes sense here. LGTM

Small note: I expect you can't put it lower than 2, since you hit the expected, but undocumented behavior here https://github.com/cortexproject/cortex/issues/4654

jhoblitt:
I believe I saw a blurb or code comment that the logic is n/2 + 1.  If that's the case, I'm guessing that 3/2 is being rounded up to 2 so the behavior may be as desired when n > 3.

jhoblitt:
I'm going to revert this.  Something odd is going on with both antu and ayekan now.

```
msg="Failed to send batch, retrying" err="server returned HTTP status 500 Internal Server Error: at least 2 live replicas required, could only find 1
```

I tried restarting all pods, which seems to be required when routing/ring changes are made sometimes, but that didn't seem to fix it.  This was working on ruka yesterday but ruka is the process of being rebuilt from scratch today.

PR #479, opened 2024-06-07
(fleet/alertbit) deploy to ayekan

fbegyn:
https://github.com/o11ydev/alertbit allows for configuring the alertbit instance as http output in Fluentbit and forwards the logs to alertmanager. The meaning is that pipelines with logs can be split out in alertmanager and those logs trigger notifications in an application.


fleet/lib/alertbit/fleet.yaml
```diff
@@ -0,0 +1,15 @@
...
+  bundle: *name
+helm:
+  chart: https://o11ydev.github.io/helm-charts
+  releaseName: *name
+  version: v0.1.0
+  valuesFiles:
+    - values.yaml
+  waitForJobs: true
+  timeoutSeconds: 900
+  atomic: false
```
jhoblitt, line 15: This is the default and we've stopped putting it in the `fleet.yaml`.

fleet/lib/alertbit/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+    alertmanagers:
+      - static_configs:
+          - targets:
+              - "alertmanager.${ .ClusterName }.${ .ClusterLabels.site }.lsst.org"
+
+ingress:
+  enabled: false
+
+resources:
+  limits:
```
jhoblitt, line 15: Please set a cpu limit.

jhoblitt (changes requested):

fleet/lib/alertbit/fleet.yaml
```diff
@@ -0,0 +1,15 @@
...
+labels:
+  bundle: *name
+helm:
+  chart: https://o11ydev.github.io/helm-charts
+  releaseName: *name
+  version: v0.1.0
+  valuesFiles:
+    - values.yaml
+  waitForJobs: true
+  timeoutSeconds: 900
```
jhoblitt, line 14: Does this service take 15 mins to roll out?

fbegyn:
Alertbit directly talks to alertmanager, so Prometheus is quite unaware of alertbit. Alertbit in itself exposes some metrics relating to the performance and in-flight alerts, but itself also holds no state of which alerts are being send to it.

What kind of PrometheusRule would you expect from it? (aside from tracking the alertbit API performance.

jhoblitt:
Aww, so there will be no prometheusrule at all, I understand now.

PR #484, opened 2024-06-07
(fleet/mimir) update README to work with ceph 18.x

jhoblitt:
It seems that radosgw-admin commands now require the realm and zonegroup to be specified even when the name of the zone would be non-ambiguous.


cbarria (approved):
i had to deal with this just now for the s3-butler

PR #487, opened 2024-06-11
(rancher.*) convert onepassword-connect install to be credentials only

jhoblitt:
Fleet will install / upgrade the onepassword-connect chart going forward.  As the credentials are being installed "manually", they will not be owned by the currently installed chart release and will not be uninstalled when the chart release is upgraded by fleet.


cbarria (approved):
Thanks so much! Looks Great 👍🏻

PR #489, opened 2024-06-12
(fleet/kube-prometheus-stack) use a ceph backed pvc for prometheus data

jhoblitt:
On clusters which have ceph configured to provide pvcs, this will prevent the prometheus tsdb from growing unbounded in the ephemeral disk space of the node on which the sts is scheduled.


fbegyn (approved):
LGTM

Some future work here then would be to add some alerting on the PVC/ceph quota for Prometheus containers specifically. Prometheus is very sensitive on storage space and a full storage for Prometheus can have fatal consequences resulting in data loss (the same also applies for the upstream long term storage nodes in regards to the temporary local TSDB).

PR #504, opened 2024-06-21
Move fluentbit deployment to fleet and helm

fbegyn:
Deploy fluent bit through helm charts and fleet. Also split up the configs for kube and external ingestation, since some clusters will only need the kube logs shipped.

This currently uses the `prom_cluster` label to distinguish which cluster is sending which logs. These will also be used in further dashboards being made available. In the future this label might change, `prom_cluster` might be a weird term since it's not Prometheus sending this data. But by using this label we have a correlation with the metric labels that can be used in searching for information.

Tested and confirmed working in `ayekan` after shaking the Opensearch nodes back into working action.


fbegyn:
@jhoblitt I might need you're input on how to best handle https://github.com/lsst-it/k8s-cookbook/pull/504/files#diff-0a6c1c007a9df1475a5e46c866ee3107054596aef7719ad66928773f3bf43bcdR44 here. This creates the services that is a `LoadBalancer` type since jamming all the possible log TCP traffic through the ingress sounds like a bad idea and it also allows for the UDP traffic. For other clusters this can be dynamically assigned, but that allows for misalignment of the load balancer IP and the DNS name used.

I'm think and overlay should probably to the heavy lifting here and an IP from the load balancer is hard assigned.

jhoblitt:
I haven't looked at this yet but, FYI, I started testing fluent-operator on kueyen several weeks ago. https://github.com/lsst-it/k8s-cookbook/compare/master...IT-5176/fluent-client

jhoblitt:
> @jhoblitt I might need you're input on how to best handle https://github.com/lsst-it/k8s-cookbook/pull/504/files#diff-0a6c1c007a9df1475a5e46c866ee3107054596aef7719ad66928773f3bf43bcdR44 here. This creates the services that is a `LoadBalancer` type since jamming all the possible log TCP traffic through the ingress sounds like a bad idea and it also allows for the UDP traffic. For other clusters this can be dynamically assigned, but that allows for misalignment of the load balancer IP and the DNS name used.
> 
> I'm think and overlay should probably to the heavy lifting here and an IP from the load balancer is hard assigned.

The way we have been handling this is setting an annotation in an overlay that sets the specific IP address.

jhoblitt (changes requested):
I have been waiting for CI to pass before reviewing this PR.

jhoblitt:
Rebased on current master.

PR #507, opened 2024-06-21
(fleet/*) prune unnecessary bundle deps

jhoblitt:
Reducing the complexity of the dep graph will speed up convergence time.


cbarria (approved):
Cleaning everyday 👍🏻

PR #513, opened 2024-06-27
(fleet/prometheus-alerts) fix HostDown rule

jhoblitt:
The `---` yaml header was breaking the helm templating of the `prometheusrule` CR.


cbarria (approved):
Thanks so much, Looks Great 👍🏻

PR #514, opened 2024-06-27
(fleet/kube-prometheus-stack) fix prom pvc filling up

jhoblitt:
On antu, a 50GiB PVC was not large enough for prom default 15 day retention period.  To prevent this from happening we are enlarging the PVC so that it should accommodate ~1weeks worth of data even on an aggegration cluster and switch from time to a size based retention window.


cbarria (approved):
awesome, thanks! Looks Great 👍🏻

PR #518, opened 2024-06-27
(fleet/kube-prometheus-stack) disable prom pvc on chonchon

jhoblitt:
As chonchon does not have ceph rbd deployed.


cbarria (approved):
Looks Great 👍🏻

PR #523, opened 2024-07-01
(fleet/cnpg-cluster) more resource to yagan deploy

cbarria:
cnpg resources were a bit tight on the summit.
current: 750m~ CPU / 914 MEM


jhoblitt (approved):
LGTM.

PR #531, opened 2024-07-09
(fleet/local-static-provisioner) add deployment on kueyen

cbarria:
when using a container version of kubelet, we need to add that bind to get  permissions to access that path.
the labels are for nodeSelector.


dtapiacl (approved):
LGTM.

PR #533, opened 2024-07-12
(fleet/local-static-provisioner) add deployment on manke

cbarria:
need to add the local provisioner to Manke to mount Kafka there on Manke08, 09, and 10.


csilva-cl (approved):
looks good

PR #534, opened 2024-07-15
(ruka) replace ruka01

cbarria:
ruka was replaced, and now it is not using rke1 but this needs to be merged to close the loop.


dtapiacl (approved):
LGTM.

PR #571, opened 2024-08-21
(fleet/local-static-provisioner) add deployment on Yagan

cbarria:
This will add the deployment for local-static-provisioner on Yagan nodes with label `local-storage: "true"` (18, 19 & 20)


dtapiacl (approved):
Looks good.

PR #577, opened 2024-08-27
(ccs-grafana/ccs-influxdb) deploy on pillan, manke & yagan

cbarria:
This was tested on Kueyen, and it is working OK. This deployment will monitor the cameras in each site.
I already created the secrets on 1password.

### DIFF
--- a/fleet/lib/ccs-grafana/values.yaml
+++ b/fleet/lib/ccs-grafana/values.yaml
@@ -64,5 +64,4 @@ datasources:
 sidecar:
   dashboards:
     enabled: true
-    namespace: ~
     label: grafana_dashboard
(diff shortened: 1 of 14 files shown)


fleet/lib/ccs-grafana/values.yaml
```diff
@@ -51,9 +51,18 @@ datasources:
...
-        password: ${`${INFLUXDB_PASSWORD}`}
+        jsonData:
+          httpMode: GET
+        secureJsonData:
+          password: ${`${INFLUXDB_PASSWORD}`}
+
+sidecar:
+  dashboards:
+    enabled: true
+    namespace: ~
```
jhoblitt, line 67: Does this result in dashboards in all namespaces being used?  That is not desirable on at least pillan.

cbarria: do'h that was for testing, going to change it back to its own namespace, but I think the default is like that and this can be removed.

jhoblitt (approved):
LGTM.

PR #581, opened 2024-09-03
(fleet/rook-ceph-conf) add rsphome export for manke

cbarria:
This ticket was developed on another branch that never got merged(pre-fleet), so I noticed this while cleaning old branches.
this is the same NFS but for Fleet.

need help checking this line `ceph nfs export create cephfs rsphome /rsphome rsphome`
not sure if it is the correct creation path (if it matches the current one)
(I think it is good, but it would be great if someone else could check this, as this is already in production)

### DIFF
--- a/fleet/lib/rook-ceph-conf/charts/manke/templates/cephnfs-rsphome.yaml
+++ b/fleet/lib/rook-ceph-conf/charts/manke/templates/cephnfs-rsphome.yaml
@@ -1,0 +1,85 @@
+---
+apiVersion: ceph.rook.io/v1
+kind: CephFilesystem
+metadata:
+  name: rsphome
+  namespace: rook-ceph
+spec:
+  metadataPool:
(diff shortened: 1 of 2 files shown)


fleet/lib/rook-ceph-conf/charts/manke/templates/cephnfs-rsphome.yaml
```diff
@@ -0,0 +1,85 @@
...
+      failureDomain: host
+      replicated:
+        size: 3
+      quotas:
+        maxSize: 250Gi
+    - name: ec
+      failureDomain: host
+      erasureCoded:
+        dataChunks: 6
+        codingChunks: 3
```
jhoblitt, line 25: This seems large for a 10 node cluster with fault tolerance of 2.  6+2 is probably safer. I wonder if 5+3 is a good idea?

PR #585, opened 2024-09-03
(*) enable fleet correctDrift for all production clusters

jhoblitt:
This leaves correctDrift disabled on these dev clusters:

- ayekan
- kueyen
- namkueyen


jhoblitt:
Reopened as https://github.com/lsst-it/k8s-cookbook/pull/585 with the jira slug fixed in the branch name.

PR #606, opened 2024-09-27
(fleet/*) set helm.timeoutSeconds for all bundles

jhoblitt:
It appears (???) that setting `helm.timeoutSeconds` may help resolve the "deadlocking" that has been observed where fleet never retries a bundle deployment.


dtapiacl (approved):
LGTM.

PR #622, opened 2024-10-07
(site/ls) bump rke to v1.29.8-rancher1-1

cbarria:
October upgrades on base, `rke client` bump to `v1.6.2` so rancher can go to `v1.29.8`

this PR goes awesome with this other: https://github.com/lsst-it/lsst-control/pull/1491


dtapiacl (approved):
lgtm.

PR #627, opened 2024-10-09
(rook-ceph/konkong) add s3 policy for s3-butler and README.md

cbarria:
this are the policies and the instructions on how to execute ticket `IT-5301`
this most probably is going to be asked to be replicated, so don't forget about this commits.

### DIFF
--- a/konkong/rook-ceph/s3/README.md
+++ b/konkong/rook-ceph/s3/README.md
@@ -1,0 +1,44 @@
+# This Policies need to be applied in the corresponding buckets for the users to grab permissions
+
+## Create the Users
+
+```bash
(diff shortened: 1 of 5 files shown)


gseriche (approved):
Is the flag --access-key --secret-key necessary? 
I understand the endpoint is Ceph.

cbarria:
> Is the flag --access-key --secret-key necessary? I understand the endpoint is Ceph.

i just removed the `KEY` and `SECRET` to show this on public

jhoblitt (approved):
This looks reasonable.  I will add the `oods-*` accounts to https://github.com/lsst-it/k8s-cookbook/pull/623

PR #637, opened 2024-10-21
(ruka/loki) Add Loki user to Ruka Ceph object store

gseriche:


cbarria (changes requested):
- Commit message is not Present Imperative.
- This user doesn't push credentials to 1password like the other users in the folder? is this correct?

cbarria (changes requested):
dunno why i got re-requested when i already reviewed this, but it is still missing lots of stuff.

csilva-cl:
> dunno why i got re-requested when i already reviewed this, but it is still missing lots of stuff.

I added you the review list, didn't mean to trigger a review, given the comment you left.

gseriche:
> * Commit message is not Present Imperative.
> * This user doesn't push credentials to 1password like the other users in the folder? is this correct?

Yes, We talked with @jhoblitt  about that; the credentials don't need to be stored in 1Password: 
>  The end users do not need access to the Loki bucket(s).  Also, Loki will be deployed in the same cluster as RGW, so pushing the credentials out to 1Pass is unnecessary.

dtapiacl (approved):
LGTM.

PR #640, opened 2024-10-22
(ruka/loki,promtail) Add observability fleet on Ruka

gseriche:
The value of the secret will be included after the PR https://github.com/lsst-it/k8s-cookbook/pull/637


fleet/s/dev/c/ruka/observability/fleet.yaml
```diff
@@ -0,0 +1,14 @@
+defaultNamespace: monitoring
```
jhoblitt, line 1: Lets install loki into a ns named `loki` to be consistent with the mimir deployment.

fleet/s/dev/c/ruka/observability/fleet.yaml
```diff
@@ -0,0 +1,14 @@
...
+helmCharts:
```
jhoblitt, line 2: This is not a valid key in a `fleet.yaml` file. https://fleet.rancher.io/ref-fleet-yaml

fleet/s/dev/c/ruka/observability/loki/values.yaml
```diff
@@ -0,0 +1,52 @@
+# loki/values.yaml
+config:
```
jhoblitt, line 2: The only config I see in the values.yaml is `loki.confg`. https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L127

fleet/s/dev/c/ruka/observability/fleet.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - name: loki
+    repo: https://grafana.github.io/helm-charts
+    chart: loki
+    version: 5.8.4
+    valuesFiles:
+      - loki/values.yaml
+  - name: promtail
```
jhoblitt, line 9: fleet does not support multiple charts in the same bundle.

jhoblitt (changes requested):
Please test this branch on ruka before re-requesting review.

PR #642, opened 2024-10-22
(ruka) change loki user on the clustersecretstore

gseriche:
Fix clustersecretstore for user Loki


cbarria (changes requested):
has an extra commit grab from somewhere and yamllint is not passing

PR #663, opened 2024-10-30
(fleet/ingress-nginx) fix ingressClassName support

jhoblitt:
Migrating from ingress class annotations to the `ing.spec.ingressClassName` field caused the multiple ingress controllers to start fighting over ing resources due to misconfiguration.


cbarria (approved):
Looks Great 👍🏻

PR #664, opened 2024-10-30
(fleet/rook-ceph-cluster) allow ceph topic configuration without TLS

jhoblitt:
Ceph, by default, does not allow http basic auth to be set in the kafka URI unless rgw TLS is enabled.  As we terminate TLS with the ingress controller, and the communication between the ingress controller and rgw is HTTP(no-s), we need to disable this sanity check.


cbarria (approved):
Thanks so much for this! looks great! 👍🏻

PR #668, opened 2024-11-01
(fleet/rook-ceph-conf) rm usage of cephnfs.spec.rados.namespace

jhoblitt:
This field is deprecated and has no effect.

```
 ~ $ k explain cephnfs.spec.rados.namespace
GROUP:      ceph.rook.io
KIND:       CephNFS
VERSION:    v1

FIELD: namespace <string>

DESCRIPTION:
    The namespace inside the Ceph pool (set by 'pool') where shared NFS-Ganesha
    config is stored.
    This setting is deprecated as it is internally set to the name of the
    CephNFS.
    

```


cbarria (approved):
Looks Great 👍🏻

PR #674, opened 2024-11-05
(fleet/multus) add configuration for ra-lhn on kueyen and yagan with demos

cbarria:
static configuration for 1 IP address that the Rapid Analysis will use to reach the USDF (tested on Kueyen working)
still need to add the CNI Plugin...

### DIFF
--- a/fleet/lib/multus-demo/fleet.yaml
+++ b/fleet/lib/multus-demo/fleet.yaml
@@ -14,6 +14,7 @@ targetCustomizations:
       values:
         networks:
           - dds
+          - ra-lhn
   - name: ruka
     clusterName: ruka
     helm:
@@ -47,3 +48,4 @@ targetCustomizations:
         networks:
           - dds
           - lhn
+          - ra-lhn-demo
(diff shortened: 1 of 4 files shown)


jhoblitt (changes requested):
Since this is a new mode for a NAD, this should be added to a test cluster like ruka as well.

dtapiacl (approved):
LGTM dude, thanks for sharing now it makes sense to me.

fleet/lib/multus-demo/fleet.yaml
```diff
@@ -47,3 +48,4 @@ targetCustomizations:
         networks:
           - dds
           - lhn
+          - ra-lhn
```
jhoblitt, line 51: Isn't this going to cause an IP address conflict on yagan if a user pod consumes the net-attach-def?

jhoblitt (changes requested):

fleet/lib/multus-demo/fleet.yaml
cbarria: yes, it is going to! 😱  there is only 1 IP address, should i leave the example just in ruka for testing purposes?

cbarria: created a demo net-attch-def and used that with the next IP in the segment

jhoblitt (approved):
LGTM.

PR #688, opened 2024-11-12
(fleet/cnpg-cluster) add max_slot_wal_keep_size

cbarria:
logical replication does need to set a max size, because the default here is `-1` which means `NO Limit` so, 10GB as half the size the database should be good.


csilva-cl (approved):
looks good. 

I suppose the "" in the max_connection is normalization, since it was already working

dtapiacl (approved):
LGTM. Thank you for taking care of that.

PR #696, opened 2024-11-20
(lfa/policies) add Lifecycle 30 days to LFA in the Summit

cbarria:
this was authorized and re-asked in the CAP Meeting on Nov 19th... according to KT, the LFA replication at the USDF only gets 1 day behind at worst scenario, so 30 days is good for the replication.
this is the last LFA without LifeCycle.


jhoblitt (approved):
LGTM.

PR #715, opened 2024-11-27
(rook-ceph/konkong) add calib user/bucket/quota/policies

cbarria:
Calibrations need to go inside their own bucket with his own user. (`calib`)
quota needs to be set at `4TB`
users:
- `butler`
- `oods-lsstcam`
- `oods-latiss`
need READ Permissions. (policy done)

README.md updated


jhoblitt (commented):
Is there a reason `cephobjectstoreuser` can't be used?

cbarria:
> Is there a reason `cephobjectstoreuser` can't be used?

i took the advice and did the CRD.

jhoblitt (approved):

PR #716, opened 2024-11-27
(konkong/s3-butler) add Lifecycle 90 days to buckets

cbarria:
all the rubinobs buckets (EXCEPT the calibration bucket) inside the s3-butler need a 90 days safety lifecycle (the butler has a 30 days auto cleaning but this is to be sure the storage doesn't get full).


jhoblitt (approved):
LGTM.

PR #718, opened 2024-11-27
(fleet/rook-ceph-conf) buff quota for olly on konkong

cbarria:
the `olly` pool was full again, we are still using it to monitor `kafka stuffs`


dtapiacl (approved):
LGTM.

PR #729, opened 2024-12-17
(observability) add targets to blackbox exporter

csilva-cl:
add http targets to scrape ssl and response time data


gseriche (approved):
LGTM

PR #740, opened 2025-01-22
S3 Butler for Summit

cbarria:
this contains the `users`, `bucket creation`, `quotas`, `policies` and `life-cycles` for all the `s3-butler` storage in the summit.


jhoblitt (approved):
At a glance, this looks reasonable.

PR #742, opened 2025-01-24
(site/dev) Add ip to metallb on Kueyen for fluentbit external

gseriche:
Add the IP 139.229.134.90 to metallb.
Add two files to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ creds.sh
 secret.yaml
 template.yaml
 **/snmp-generator/mibs
+.DS_Store
+*\ [2-9].sh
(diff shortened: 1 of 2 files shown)


jhoblitt (changes requested):
ipam shows 139.229.134.90 as in use.  Please determine if there is an assignment conflict.

gseriche:
> ipam shows 139.229.134.90 as in use. Please determine if there is an assignment conflict.

I change that assignment.

image

dtapiacl (approved):
LGTM. Thanks for updating IPAM.

PR #744, opened 2025-01-27
(elqui/rook-ceph-conf) set maxBuckets 2 users

cbarria:
users `comcam`, `latiss` and `lsstcam` need to own their respective `raw` and `butler` buckets, so `maxBuckets` needs to be set to `2`. (all other users are managed by policies.)


gseriche (approved):
LGTM

PR #746, opened 2025-01-29
(fleet/rook/ceph/conf) add mimir s3 account to kueyen

gseriche:
Mimir requires a S3 bucket to store the tsdb data.


cbarria (commented):
did you test this resources manually (or with fleet) on `Kueyen` to be sure they are good before merging?

gseriche:
> Did you test these resources manually (or with fleet) on `Kueyen` to be sure they are good before merging?

The resource was created successfully; I tested the user/pass and made some buckets there. Currently, Mimir is working on that.

PR #755, opened 2025-02-12
(kueyen) Add nodes 04 and 05 to the cluster

gseriche:


dtapiacl (approved):
This looks good to me, thanks.

gseriche:
> I don't object to this but it would be better to convert the cluster to rke2 instead of continuing to update the rke config. rke[1] will be EOL on 2025-07-31.

I agree with this, @jhoblitt. After I finish my part with the observability, I'll plan to move to RKE2

PR #773, opened 2025-02-26
(kueyen/mimir) Add remotewrite to Kueyen

gseriche:


jhoblitt (changes requested):
This would cause pillan, antu, ruka, and ayekan, which are already configured as "aggegrators",  to send all metrics kueyen.  This would result in kueyen receiving all metrics at least 4 times.

jhoblitt:
Actually, only 3 times as pillan can't access kueyen. It would kill the logs on pillan with remotewrite errors though.

csilva-cl:
you need to add kueyen to the cl-remotewrite, if not is going to be pushed to all clusters.

jhoblitt (approved):

PR #774, opened 2025-02-26
(mimir) remove unused remotewrite

gseriche:


jhoblitt (commented):
Why would we remove these?  mimir is still running on those clusters.

jhoblitt:
> > Why would we remove these? mimir is still running on those clusters.
> 
> Ayekan will be rebuilt. Ruka is not needed. @csilva-cl, your comments, please.

`ruka` is currently the only test environment for `antu` with a fleetized configuration.  `ayekan` will presumably be rebuilt into another general purpose test cluster, which will include all of our production services including mimir.

csilva-cl:
I agree on ruka, but I dont think we will deploy mimir on ayekan. We already have kueyen which should server as the dev. of antu. So 2 dev clusters for antu is probably enough

jhoblitt (approved):

PR #784, opened 2025-03-03
(obs/dashbaords) fix datasources for all dashboards

csilva-cl:
dashboards had the wrong data source; this has been solved by setting the "datasource" variable.


cbarria (approved):
Looks Great 👍🏻

PR #789, opened 2025-03-06
(fleet/cnpg-cluster) bump manke resources

cbarria:
cnpg in manke was getting killed by the OOM, so resources were bump.

### DIFF
--- a/fleet/lib/cnpg-cluster/overlays/manke/cluster-cnpg-cluster.yaml
+++ b/fleet/lib/cnpg-cluster/overlays/manke/cluster-cnpg-cluster.yaml
@@ -35,8 +35,8 @@ spec:
 
   resources:
     limits:
-      cpu: "1"
-      memory: 1Gi
+      cpu: "2"
+      memory: 4Gi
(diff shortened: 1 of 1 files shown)


ktlim:
I would suggest making this similar to, or at worst slightly smaller than (not half of), the configuration at the Summit, which is:
```
  resources:
    limits:
      cpu: '4'
      memory: 8Gi
    requests:
      cpu: '2'
      memory: 4Gi
```

gseriche (approved):
lgtm

dtapiacl (approved):
Looks reasonable to start with, if more resources are needed we can bump it.

cbarria:
> I would suggest making this similar to, or at worst slightly smaller than (not half of), the configuration at the Summit, which is:
> 
> ```
>   resources:
>     limits:
>       cpu: '4'
>       memory: 8Gi
>     requests:
>       cpu: '2'
>       memory: 4Gi
> ```

we are over selling resources everywhere, if more is needed, we can bump it again.

PR #793, opened 2025-03-19
(fleet/kube-prometheus-stack/overlays/antu) Update timeout and interval

badenerb:


csilva-cl:
I wonder if we should increase the scrape rate too, given the pdus take ~40s to scrape

jhoblitt:
Is this testing as working?  I vaguely recall testing really high values in the past, as in several minutes, and was still seeing timeouts but there have been several updates to the snmp exporter since then.

csilva-cl:
probably worth testing, but 5 minutes seems like an overkill, 1m or 2m would be sufficient as scrape rate

badenerb:
I updated the interval to 2m instead of 5

jhoblitt (approved):

PR #814, opened 2025-03-26
(fleet/*) rm usage of nginx.ingress.kubernetes.io/server-snippet

jhoblitt:
This annotation was being used to disable TLS cert validation but is unnecessary when the backend is explicitly HTTP.


cbarria (approved):
Looks Great 👍🏻

PR #820, opened 2025-03-31
(multus/yagan) add route to embargo on NAD lhn-ra

cbarria:
this fixes the route needed to reach the embargo s3 on multus.


gseriche (approved):
lgtm

dtapiacl (approved):
Looks great thanks !

jhoblitt (approved):
I'm not thrilled about these static net-attach-defs with duplicate settings but I'm OK with this update as long as no more static defs are created.

PR #825, opened 2025-04-01
(antu) Add cephobjectstore LFA

gseriche:


jhoblitt (changes requested):
What is the motivation for adding a second rgw instance to antu?  Is there a plan to migrate the existing mimir data over from the existing o11y rgw instance to the new lfa instance and consolidating to a single instance or is there a reason to run two completely independent rgw instances?

Also note that the pg_num values in this PR are too large for single rgw instance on antu and definitely too large to coexist with the existing o11y rgw instance.

gseriche:
> What is the motivation for adding a second RGW instance to Antu? Is there a plan to migrate the existing mimir data from the existing o11y rgw instance to the new lfa instance and consolidate to a single instance, or is there a reason to run two completely independent rgw instances?
> 
> Also note that the pg_num values in this PR are too large for a single RGW instance on antu and too large to coexist with the existing o11y RGW instance.

It was a mini discussion with @cbarria on the channel. He suggested always making your deployments standalone unless they are meant to be glued together.
Of course, I don't want to need the O11y store forever. I'm just standardizing what is made of Kueyen—the same LFA objectstore for Loki and Mimir.

In summary, creating a new store for Loki would be okay. Once this is done, I can create a Loki user for that store. That makes sense to me at that moment. 
Now, about the pg_num, I saw it in Elqui, the creation of that store. I will receive suggestions on it.

PR #826, opened 2025-04-01
(antu) Add cephobjectstoreuser loki

gseriche:
Pending PR: #825


jhoblitt (changes requested):
Our typical workflow is that an application deployment is refined in the dev env, then fleetized in dev, and then that configuration is copied forward from dev into production.  I have not seen instructions on how to view a demo loki deployment in dev or seen/reviewed a fleetized configuration for dev. Please open a PR(s) to deployment loki in dev. I would then like the dev cluster to be reinstalled from scratch to verify that the deployment is complete.

I do not want to do a piecemeal deployment to production for loki. I would like to see all of the necessary components deployed to antu in a single PR after the fleet deployment is known to be working dev.

dtapiacl (approved):
These changes look reasonable to me but I would test this in you workspace in Antu before merging.

cbarria (approved):
Looks reasonable.

PR #828, opened 2025-04-01
(rke2/rancher) Upgrade rancher.dev to RKE2

cbarria:
since the rancher deployment is not automated, we need to setup basic deployments in order to reach `fleets` automation.
- `metalLB` (required for ingress)
- `Ingress` (required for open-connect/external-secrets)
- `cert-manager` (required for open-connect/external-secrets)
- `external-secrets` (required for deploying secrets)
- `open-connect` (required for 1password)
- `rancher-backup` operator (all the adobe + access to aws)
- `rancher` (all the adobe)


jhoblitt (commented):
This seems like a lot of components to boot strap rancher / fleet. I don't believe that external-secrets (other than injecting the css secret), open-connect (other than the connect secret), or rancher-backup should be needed.  Once rancher is installed, it should be able to install the remaining services.

jhoblitt:
> > This seems like a lot of components to boot strap rancher / fleet. I don't believe that external-secrets (other than injecting the css secret), open-connect (other than the connect secret), or rancher-backup should be needed. Once rancher is installed, it should be able to install the remaining services.
> 
> the problem is the backup operator needs to go before the rancher in order to restore, and this needs an external secret to grab the restore from AWS, so it is all tied up... but yes, once fleet gets to run, the deployments will be straighten. @jhoblitt

Is isn't possible to restore the rancher state on top of an existing deployment?

cbarria:
> > > This seems like a lot of components to boot strap rancher / fleet. I don't believe that external-secrets (other than injecting the css secret), open-connect (other than the connect secret), or rancher-backup should be needed. Once rancher is installed, it should be able to install the remaining services.
> > 
> > 
> > the problem is the backup operator needs to go before the rancher in order to restore, and this needs an external secret to grab the restore from AWS, so it is all tied up... but yes, once fleet gets to run, the deployments will be straighten. @jhoblitt
> 
> Is isn't possible to restore the rancher state on top of an existing deployment?

yes, but no when provisioning the cluster from scratch, as it gets considered Migrating into another cluster. (RKE1->RKE2 issue, there are some records colliding)

jhoblitt (commented):
Could we split the rancher upgrade out of this and deploy it immediately rather than waiting for rke2 conversion?

gseriche (approved):
lgtm

PR #834, opened 2025-04-07
(antu) adjust timeout and scrape interval

csilva-cl:
pdu timeout and scrape interval reduced
network timeout reduced and scrape increased.


dtapiacl (approved):
LGTM.

PR #844, opened 2025-04-10
(luan) RKE1 --> RKE2 Migration

dtapiacl:


gseriche (approved):
lgtm

fleet/lib/rook-ceph-cluster/overlays/luan/values.yaml
```diff
@@ -1,33 +1,45 @@
 ---
 cephClusterSpec:
   mon:
-    count: 5
+    count: 3
```
jhoblitt, line 4: We probably want all nodes in a 5 node cluster to be a control-plane node and have a mon.

fleet/lib/rook-ceph-cluster/overlays/luan/values.yaml
```diff
@@ -1,33 +1,45 @@
...
+          - name: /dev/disk/by-id/ata-WDC_WDS400T2B0A-00SM50_214227800496
       - name: luan05
         devices:
-          - name: sda
-          - name: sdb
+          - name: /dev/disk/by-id/ata-WDC_WDS400T2B0A-00SM50_214227800262
+          - name: /dev/disk/by-id/ata-WDC_WDS400T2B0A-00SM50_214227800511
 
 cephBlockPools:
   - name: replicapool
```
jhoblitt, line 45: We want to set at least deletion protection on `replicapool`.
```
      parameters:
        nodelete: "true"
        nosizechange: "true"
```

jhoblitt (changes requested):
Additionally, the cephnfs should be removed from luan as we don't want any nfs exports for this cluster.  The `cnpg-system` bundle can be removed as well.

jhoblitt (approved):
LGTM.

PR #847, opened 2025-04-10
(fleet/snmp-exporter) update pdu fqdn

csilva-cl:
- updated wrong fqdn
- removed repeated ip addresses


jhoblitt (approved):
This looks plausible to me but I haven't tried to check dns records or ipam for any of the IPs.

PR #850, opened 2025-04-14
(fleet/cnpg-cluster) boost yagan resources and buffers.

cbarria:
-  `shared_buffers: 4GB` recommended 25% of the total RAM (more buffer, less disk writing!)
-  `work_mem: 64MB` bumping from 4MB (more memory to do `JOIN`, `SORT`, etc.)
-  `effective_cache_size: 12GB` (tells the planner it has more memory to assign ~75% of total)
-  `idle_session_timeout: 1h` (reducing the timeout for idle because of poor usage of sockets on Code connecting to DB)

### DIFF
--- a/fleet/lib/cnpg-cluster/overlays/manke/cluster-cnpg-cluster.yaml
+++ b/fleet/lib/cnpg-cluster/overlays/manke/cluster-cnpg-cluster.yaml
@@ -13,8 +13,10 @@ spec:
   postgresql:
     parameters:
       max_connections: "500"
-      shared_buffers: 256MB
-      idle_session_timeout: 4h
+      shared_buffers: 2GB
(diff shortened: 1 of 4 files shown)


csilva-cl (approved):
looks good

jhoblitt (commented):
Do we need to increase this in TTS/BTS as well?  Why is there resource contention at the summit but not in either of the test stands?

cbarria:
> Do we need to increase this in TTS/BTS as well? Why is there resource contention at the summit but not in either of the test stands?

you are right, i will update all the config with each site spec.

jhoblitt (approved):
LGTM.

dtapiacl (approved):
LGTM.

PR #851, opened 2025-04-15
(fleet/cnpg-cluster) add more slots and memory to yagan.

cbarria:
replication es getting way behind (almost a day according to grafana) so we need more replication slots.
counting for 3 replicas:
- 3 Wal Senders.
- 3 Autovaccum
- 20 logical replication
- ~6 buffer (2 ea)
Total: 32... so i bumped `max_worker_processes: "40"` from `32`

also, from testing last night, `16Gi` of memory was on the edge around `93%` use, so bump to `32Gi` should put us on `~50%`


dtapiacl (approved):
Looks great.

gseriche (approved):
lgtm

PR #876, opened 2025-04-25
(fleet/cnpg-cluster) bump max_connection and bouncer

cbarria:
- bump max_connections: 1000 (max recorded 752)
- adjust bouncer to those 1000 connections and prioritize butler DB
- idle disconnect bumped to 20mins. (was too short before)


csilva-cl (approved):
looks good

PR #877, opened 2025-04-25
(fleet/rook-ceph-cluster) replace yagan01 disk

cbarria:
The disk died, we cut the power for a few mins didn't come back. (@badenerb  found another exact disk 👏🏻 )


dtapiacl (approved):
Looks great.

dtapiacl:
This should address the OSD1 issues.

PR #878, opened 2025-04-25
(fleet/cert-manager) remove cert-manager and crds

cbarria:
- ArgoCD took cert-manager con Pillan, Manke and Yagan.
- cert-manager-conf is still creating the cluster-issuer


dtapiacl (approved):
Looks good.

jhoblitt (changes requested):
~ s/con/from/

cbarria:
> ~ s/con/from/

i think i meant to write "on" and took an extra "c" with the space bar, fixed.

jhoblitt (approved):

PR #879, opened 2025-04-28
(site/ls) bump rke to v1.31.7-rancher1-1

dtapiacl:
April 2025 OS upgrades include the following:

RKE Client is bumped to v1.7.6 to support K8S version v1.31.7-rancher-1-1

This version bump will be effective as of tomorrow.

### DIFF
--- a/antu/rke/cluster.yml
+++ b/antu/rke/cluster.yml
@@ -45,6 +45,6 @@ network:
   plugin: canal
 ssh_key_path: ~/.ssh/id_rsa
 ignore_docker_version: true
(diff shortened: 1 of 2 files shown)


jhoblitt (changes requested):
image
There is a trailing `)` in both the PR title and the commit message.

There is 1 too many periods in both `v.1.7.6`  and `v.1.31.7-rancher-1-1`.

cbarria (approved):
Looks Great, but fix the typo xD

jhoblitt (approved):
LGTM.

PR #888, opened 2025-04-28
(fleet/cnpg-cluster) fix config pooler

cbarria:
- pgbouncer on cnpg doesnt allow settings for individual databases, so config was changed to affect all databases.
- removed the timeouts, only left the 1 hour (same as the database)


csilva-cl (approved):
looks good

PR #906, opened 2025-05-02
(fleet/kube-prometheus-stack) lower request/limits on rancher clusters

cbarria:
- this lowers the request/limits on kube-prometheus-stack for the rancher clusters.
*this cannot go without merging first* https://github.com/lsst-it/k8s-cookbook/pull/912

### DIFF
--- a/fleet/lib/kube-prometheus-stack/fleet.yaml
+++ b/fleet/lib/kube-prometheus-stack/fleet.yaml
@@ -83,9 +83,10 @@ targetCustomizations:
         - key: management.cattle.io/cluster-display-name
           operator: In
           values:
-            - konkong
             - local
     helm:
+      valuesFiles:
+        - low-resource/values.yaml
       values:
         prometheus:
           prometheusSpec:
@@ -121,6 +122,8 @@ targetCustomizations:
           values:
             - local
     helm:
+      valuesFiles:
+        - low-resource/values.yaml
       values:
         prometheus:
           prometheusSpec:
--- a/fleet/lib/kube-prometheus-stack/low-resource/values.yaml
+++ b/fleet/lib/kube-prometheus-stack/low-resource/values.yaml
@@ -1,0 +1,10 @@
+---
+prometheus:
+  prometheusSpec:
+    resources:
+      limits:
+        cpu: 500m
+        memory: 2Gi
+      requests:
+        cpu: 100m
+        memory: 1Gi


gseriche (approved):
lgtm

fleet/lib/kube-prometheus-stack/fleet.yaml
```diff
@@ -136,3 +136,21 @@ targetCustomizations:
         prometheus:
           prometheusSpec:
             remoteWrite: *tu-remotewrite
+  - name: local-low-prometheus
+    clusterSelector:
+      matchExpressions:
+        - key: management.cattle.io/cluster-display-name
+          operator: In
+          values:
+            - local
```
jhoblitt, line 145: This target will never match because there is already an explicit name match on line 83: https://github.com/lsst-it/k8s-cookbook/pull/906/files#diff-d143cb3f98049980638fd7a43dc711a0f0994c6b661e879d9215d9a663329322R83 

and then all of the dev/ls/cp clusters will match https://github.com/lsst-it/k8s-cookbook/pull/906/files#diff-d143cb3f98049980638fd7a43dc711a0f0994c6b661e879d9215d9a663329322R83

jhoblitt (changes requested):

jhoblitt (changes requested):
`s/kue/kube/`

fleet/lib/kube-prometheus-stack/fleet.yaml
cbarria: i needed more knowledge, so now i know that multiple `matchExpressions` act like  `AND` and must match all the `Keys` in the selector. 💪🏻

fleet/lib/kube-prometheus-stack/fleet.yaml
```diff
@@ -88,6 +87,13 @@ targetCustomizations:
               - url: https://mimir.ruka.dev.lsst.org/api/v1/push
               - url: https://mimir.antu.ls.lsst.org/api/v1/push
               - url: https://mimir.kueyen.dev.lsst.org/api/v1/push
+            resources:
```
jhoblitt, line 90: Maybe this would be better in a `valuesFile` as I don't think there is a reason to have tu & cl have different values?  That's less of an opportunity for them to drift out of sync in the future.

jhoblitt (approved):
LGTM.

fleet/lib/kube-prometheus-stack/fleet.yaml
cbarria: changed! great idea

PR #911, opened 2025-05-05
(fleet/lib/ccs-influxdb/values.yaml) bump PVC size

badenerb:
This solves the issue of the PVC being filled, but it is a temporary solution. Max Turri will need to find the process that is filling the PVC (as seen on the Grafana Screenshot).
image (5)


dtapiacl (approved):
Looks reasonable to me. I think it wise if we just bump the size instead of deleting stuff from their database.

jhoblitt (approved):

PR #925, opened 2025-05-09
(gha) add block merge commits workflow

jhoblitt:


cbarria (approved):
thanks for setting this up!

PR #934, opened 2025-05-15
(kueyen) Add Loki to Kueyen

gseriche:
(ayekan,konkong,elqui) convert CephBucketTopic kafka configuration to use k8s secrets


csilva-cl (changes requested):
You need to add loki as a datasource to Grafana.

fleet/lib/loki/overlays/kueyen/values.yaml
```diff
@@ -0,0 +1,58 @@
+loki:
+  storage:
+    s3:
+      endpoint: http://rook-ceph-rgw-o11y.rook-ceph.svc.cluster.local:80
+
+  commonConfig:
+    storage:
```
jhoblitt, line 7: I believe this is a noop as the template has a hardcoded `storage` key, which will conflict.

```
    {{- if .Values.loki.commonConfig}}
    common:
    {{- toYaml .Values.loki.commonConfig | nindent 2}}
      storage:
      {{- include "loki.commonStorageConfig" . | nindent 4}}
    {{- end}}
```

fleet/lib/loki/overlays/kueyen/values.yaml
```diff
@@ -0,0 +1,58 @@
...
+      s3:
+        endpoint: http://rook-ceph-rgw-o11y.rook-ceph.svc.cluster.local:80
+
+  limits_config:
+    retention_period: 24h
+    retention_stream:
+      - selector: '{app=~".+"}'
+        period: 24h
```
jhoblitt, line 15: Is this redundant with a global `retention_period` set to the same value?

fleet/lib/loki/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+  auth_enabled: false
+
+  storage:
+    type: s3
+    s3:
+      access_key_id: ${`${AWS_ACCESS_KEY_ID}`}
+      secret_access_key: ${`${AWS_SECRET_ACCESS_KEY}`}
+      s3ForcePathStyle: true
+      region: o11y
+    bucketNames:
```
jhoblitt, line 18: Do these buckets need to be pre-created?

fleet/lib/loki/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+read:
+  replicas: 0
+write:
+  replicas: 0
+backend:
+  replicas: 0
+
+test:
+  enabled: true
```
jhoblitt, line 70: This is the default value and doesn't need to be set.

fleet/lib/loki/values.yaml
gseriche: I tried using the OBC, but it didn't work last time. So, I created the user and then made each bucket.

csilva-cl: that's fine for testing, but creation of buckets should be included in the PR

gseriche: Do you have any ideas on how I can create the buckets then? I tried with ObjectBucketClaim, but it didn't work; it only applies policies to the bucket. It didn't create the buckets on ceph, so I made them manually with the account.

fleet/lib/loki-pre/overlays/kueyen/externalsecret-loki-s3.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
```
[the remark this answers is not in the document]
gseriche: Based on this, should I copy the access/key from rook-ceph to the Loki namespace and call it right?

fleet/lib/loki/overlays/kueyen/values.yaml
gseriche: Done

gseriche:
> You need to add loki as a datasource to Grafana.

Done

cbarria (approved):
as long as the values used are correct, the structure looks good.

PR #938, opened 2025-05-16
(*) fix ownership of rubinobs-butler-* buckets

jhoblitt:
This is a partial re-sync of cross sight configuration.  TTS remains mostly out of sync with the other sites.  As the configuration has continued to grow, this manual re-sync'ing is increasingly tedious.  In the future, we may want to move the oods/etc. related configure into its own chart and live with comcam buckets/etc. being configured in the bts so as to minimize the cross-site differences.


cbarria (approved):
Thanks for fixing this 👍🏻

mareuter (approved):
Looks good to me.

PR #940, opened 2025-05-22
(dev) Add fluentbit to dev site.

gseriche:


jhoblitt (commented):
I'm not sure I understand the title of this PR.  The bundle being modified isn't included by any cluster.  Is the intent to deploy fluentbit on all dev clusters?

gseriche:
> I'm not sure I understand the title of this PR. The bundle being modified isn't included by any cluster. Is the intent to deploy fluentbit on all dev clusters?

Yes, @jhoblitt, and next steps should be to add fluent-bit to the ls and cp.
Fluent-bit will extract the logs from the containers and send them to Loki.

csilva-cl:
> it looks ok, but i want to confirm it was tested in dev. (I cant see the ns)

scratch that, I can see the ns

cbarria (approved):
lgtm

PR #943, opened 2025-05-27
ranchers standardize and migration of base instance.

cbarria:
took the opportunity to DRY up the cookbook on the rancher side using the template folder on most scripts.
- Add config for migrating rancher.ls into RKE2


jhoblitt (changes requested):
The overall effect of this PR LGTM but I have a quibble with a commit title.

The `()` subject of this commit doesn't make sense to me: 
image
I don't believe `cook-book` is a subject that helps locate any files in this repo.

cbarria:
> The overall effect of this PR LGTM but I have a quibble with a commit title.
> 
> The `()` subject of this commit doesn't make sense to me: image I don't believe `cook-book` is a subject that helps locate any files in this repo.

yeah, was a poor choice of words on my part, i will get that fixed lol

gseriche (approved):
Looks great to me.
The fetch-credentials.sh is standardized for all installations/migration/upgrade?

PR #947, opened 2025-06-02
(site/cp) bump rke to v1.31.8-rancher1-1

dtapiacl:
This PR goes well with: https://github.com/lsst-it/lsst-control/pull/1779


cbarria (approved):
Looks Great 👍🏻

jhoblitt (changes requested):
As rancher.cp is in the process of being upgraded to rke2, I think that can be removed this PR. Otherwise, LGTM.

PR #949, opened 2025-06-09
(fleet/cnpg-cluster) bump cnpgsphere 16.8 on pillan

cbarria:
bumping `postgres` version on the `TTS` to `16.8`


gseriche (approved):
lgtm

PR #956, opened 2025-06-16
(fleet/k8up) add k8up deployment 

dtapiacl:
1. Sets up k8up deployment on dev.
2. Tested with Nexus on ruka.
3. Adds Readme.md  with a small instruction set.


fleet/lib/k8up-post/overlays/ruka/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+  - backup-namespace.yaml
```
csilva-cl, line 2: perhaps it would be good to call it backup-daily instead of backup-namespace, so we can group all namespaces that will be backed up every day. If we need to change the schedule we can create another one. (ie backup-weekly)

csilva-cl (changes requested):

csilva-cl (approved):
looks good, thanks for the Readme.

PR #963, opened 2025-07-01
(antu) Add Loki bundle

gseriche:


cbarria (changes requested):
the name of the commit is wrong, check the history and should be something around (fleet/bundle) action in imperative present.

fleet/lib/loki-pre/overlays/antu/externalsecret-loki-s3.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: external-secrets.io/v1beta1
```
cbarria, line 2: this file is identical to the previous on kueyen, this bundle doesn't need a overlay, please remove it.

fleet/lib/loki/overlays/antu/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+queryScheduler:
+  replicas: 1
+
+indexGateway:
+  replicas: 1
+
+gateway:
+  ingress:
+    hosts:
+      - host: loki.antu.ls.lsst.org
```
cbarria, line 40: why is this here? if the default values will do the same?
`loki.${ get .ClusterLabels "management.cattle.io/cluster-display-name" }.${ .ClusterLabels.site }.lsst.org`

cbarria (changes requested):

jhoblitt (dismissed):
As loki is deploying into the same cluster as the ceph instance, it isn't necessary to sync the credentials to 1pass.  A `cephobjectstoreuser` should be used to create the credentials directly in the loki ns.

csilva-cl (approved):
looks reasonable.

PR #975, opened 2025-07-15
(fleet/cnpg-cluster) activate pgaudit on ruka

cbarria:
- activates the `pgaudit` on the `cnpg` deploy in `ruka`


cbarria:
> Is there automatic truncation or rotation of this log file?

the `pgaudit` just shoots logs from the transactions to the stout of the POD, the handling of the Logs goes on the K8s side and should be picked up by the log monitoring in the future. 👍🏻

PR #1023, opened 2025-08-05
(flee/rook-ceph-conf) add yagan k8up bucket

dtapiacl:
This creates a bucket and user to push k8up backups from Yagan onto Gaw's S3. This is also the first step towards moving k8up on to the summit for backups.


fleet/lib/rook-ceph-conf/charts/gaw/templates/obc-yagan-k8up.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: ObjectBucketClaim
+metadata:
+  name: &name yagan-k8up
+  namespace: rook-ceph
+spec:
+  bucketName: *name
+  storageClassName: k8up-backups
+  additionalConfig:
+    bucketOwner: yagan-k8up
+    bucketMaxSize: 10Ti
```
csilva-cl, line 12: yagan is much larger, I would increase bucket size according to current yagan use.

csilva-cl (changes requested):

fleet/lib/rook-ceph-conf/charts/gaw/templates/obc-yagan-k8up.yaml
dtapiacl: Yeah, i agree its on DRAFT at the moment. I had some questions regards this that I wanted to address with you because we are talking big amounts of data.

csilva-cl (approved):
looks good

PR #1084, opened 2025-08-25
(fleet/mimir) increase store_gateway pvc to 100Gi

jhoblitt:
Per https://grafana.com/docs/mimir/latest/manage/run-production-environment/planning-capacity/, disk should be ~72GiB to support our current active series size of ~5.5M.  Bumping up to 100Gi is to allow some room for growth.


csilva-cl (approved):
looks good

PR #1094, opened 2025-08-25
(fleet/mimir) tune resource limits

jhoblitt:
Per: https://grafana.com/docs/mimir/latest/manage/run-production-environment/planning-capacity/

### DIFF
--- a/fleet/lib/mimir/values.yaml
+++ b/fleet/lib/mimir/values.yaml
@@ -54,14 +54,16 @@ compactor:
     size: 100Gi
   resources:
     limits:
-      memory: 3Gi
+      cpu: 1
+      memory: 4Gi
     requests:
       cpu: 1
-      memory: 3Gi
+      memory: 4Gi
 distributor:
   replicas: 2
   resources:
     limits:
+      cpu: 12
       memory: 12Gi
     requests:
       cpu: 1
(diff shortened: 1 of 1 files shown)


fleet/lib/mimir/values.yaml
```diff
@@ -100,25 +102,37 @@ ingester:
   zoneAwareReplication:
     topologyKey: kubernetes.io/hostname
 querier:
-  replicas: 1
+  replicas: 2
```
csilva-cl, line 105: it seems 2 is the default for replicas

jhoblitt: I think we should set it explicitly as some of the components default to 1 replica for no particular reason and it seems like the chart defaults could either be normalized to not be HA or to be all HA and we don't know which (if any) will happen. E.g.: https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/values.yaml#L1946

fleet/lib/mimir/values.yaml
```diff
@@ -129,10 +143,11 @@ store_gateway:
   replicas: 3
   resources:
     limits:
-      memory: 2.5Gi
+      cpu: 10
+      memory: 10Gi
     requests:
       cpu: 1
-      memory: 2.5Gi
+      memory: 10Gi
```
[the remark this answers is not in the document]
jhoblitt: With an assumed max qps of 100.

fleet/lib/mimir/values.yaml
csilva-cl: that works for me, I just wanted to make sure that was the intention. Thank you.

csilva-cl (approved):
all good

PR #1112, opened 2025-09-03
(fleet/nexus) add nexus pg env variables dev

dtapiacl:
Adds necessary PG env variables on the Nexus instance running on Ruka. This is currently being tested on fleet Ruka and is working as expected.


dtapiacl:
> there's no url key in onepass

Its there, i just added and additional field to `nexus-postgres-creds` .
Screenshot 2025-09-04 at 13 58 14

PR #1114, opened 2025-09-04
(fleet/prometheus-alerts) add pdu alerts and squadcast-oncall

csilva-cl:


fleet/lib/kube-prometheus-stack/overlays/antu/values.yaml
gseriche, line 1: Quick question, why not LS instead of antu?

dtapiacl (approved):
LGTM

PR #1151, opened 2025-10-01
(fleet/fleet-conf) fix indention of clusterSelector fields

jhoblitt:


cbarria (approved):
Looks Great, thanks for fixing it

PR #1501, opened 2026-03-09
(ruka,pillan,manke) revert to ganesha v5.9

jhoblitt:
Due to bug(s) which break cloning git repos with git >= ~2.48.0.


csilva-cl (approved):
looks good

PR #1546, opened 2026-04-08
(fleet/k8up-post) adjust prune values

dtapiacl:
The S3 bucket (manke-k8up) filled up due to insufficient prune settings. With 7 PVCs backing up daily, the previous configuration of keepLast: 10 with a weekly prune schedule was not keeping up with snapshot accumulation. 

This keeps 14 snapshots per PVC instead of 10 across everything.


csilva-cl (approved):
looks possible.

PR #1551, opened 2026-04-09
(fleet/rook-ceph-cluster) Fix Disk IDs for Kueyen Cluster

badenerb:


dtapiacl (changes requested):
This looks good, but this is not a bug, it's just an enhancement please remove the label as its not needed.

dtapiacl (approved):
Looks good.

PR #1560, opened 2026-04-27
(fleet/metallb) bump probes delay

badenerb:


csilva-cl (changes requested):
this is a site wide change, why is this needed ?

dtapiacl (changes requested):
Looks possible but not sure why this is only affecting Kueyen. 

I suggest creating an overlay to test this only on Kueyen at: `fleet/lib/metallb/overlays/kueyen` with those values. 

You can use something like k8up deployment as an example.

badenerb:
> Looks possible but not sure why this is only affecting Kueyen.
> 
> I suggest creating an overlay to test this only on Kueyen at: `fleet/lib/metallb/overlays/kueyen` with those values.
> 
> You can use something like k8up deployment as an example.

All fixed with this suggestion, tested and working.

badenerb:
> this is a site wide change, why is this needed ?

Agreed and all fixed to be only in kueyen

csilva-cl (changes requested):
I still dont understand why is this affecting kueyen and not the rest, can you explain pls?

badenerb:
> I still dont understand why is this affecting kueyen and not the rest, can you explain pls?

I am not 100% sure why Kueyen needs more time to start up metallb controller, but my current theory is that the Dell Blades we are using are just slower because they are older (at least 5 years old), so, they need more time to spin up the pod than the default value of 10 seconds.

dtapiacl:
> I still dont understand why is this affecting kueyen and not the rest, can you explain pls?

This symptom is affecting all clusters in dev/ls at the moment. CP is not affected by this as the site has not been rolled. We tested this fix on ruka as well using the workspace setup on rancher and it seems to address the issue there. 

We suspect that this PR introduced a bug on to our cluster which included various dependencies updates that reached both dev/ls. 

Link: https://github.com/lsst-it/lsst-control/compare/cp_production...production

dtapiacl (approved):
LGTM.

csilva-cl (approved):
ok

PR #1585, opened 2026-05-11
(fleet/loki) reduce ingester replicas for dev

csilva-cl:
dev can only hold 1 replica per node, and kona is a 4 nodes cluster.


dtapiacl (commented):
I'm confused, are you removing or adding ingester replicas?

csilva-cl:
> I'm confused, are you removing or adding ingester replicas?

there can be only 1 replica per node and kona has 4 nodes. This is not to remove replicas, just to reduce from 6 ro 4

dtapiacl (approved):
LGTM.

PR #1602, opened 2026-07-02
(fleet/cnpg-cluster-new)

dtapiacl:
Deploys a parallel instance of CNPG at the summit in which we will do a complete restore of the database on to.


igonzalezcl (approved):
LGTM

PR #1608, opened 2026-07-03
(fleet/cnpg-cluster-new) adjust new cnpg cluster resources

dtapiacl:
Shrinks resources temporarily to fit the instance in the cluster that is running parallel with the production instance. Once we confirm everything works, we can retire the current instance and bump the resources once more.


badenerb (approved):
LGTM

PR #1614, opened 2026-07-06
(fleet/cnpg-cluster-new) add backup

dtapiacl:
Enables backups on the newly deployed instance of CNPG at the summit.


badenerb (approved):
lgtm

PR #1625, opened 2026-09-11
(fleet/rook-ceph-cluster) add toqui lists

dtapiacl:
Adds toqui fast disks to rook-ceph-cluster disk configuration


badenerb (approved):
lgtm, good catch

